Issue #3, opened 2020-08-08
Make the data-dir configurable. 


kmova:
The data directory used for creating the sparse file is hardcoded to "/var/csi/rawfile" in the task template. 
https://github.com/openebs/rawfile-localpv/blob/master/templates/task.yaml#L14

Similar to image_tag and image_repository, the data-dir also should be configurable via the env variable. 

Note: If you want to get started on setting up rawfile and checking how it works, you can follow the helm3 instructions in the repo or use the YAML instructions available here:

https://github.com/kmova/bootstrap/blob/master/openebs-localpv/rawfile/setup.md

jshin47:
I too would appreciate making this configurable and varied by storage class.

mhkarimi1383:
Hi @kmova 

I have made this possible in https://github.com/openebs/rawfile-localpv/pull/28 and also I have made other changes in helm chart to be more customizable :), Check it out if you want :)

thetylerwolf:
+1 on this

PR #4, opened 2020-09-30
Feature: make data dir configurable

ronaldyuwandika:
This is PR for fix this issue #3 
 
What I do:

- add support to dynamically change the default CSI driver rawfile and data directory

How to tests:

- Just change the value of  `rawfile_dir` and `data_dir` in the helm values file stored at `deploy/charts/rawfile-csi/values.yaml`

Signed-off-by: ronaldyuwandika ronald.d.yuwandika@gmail.com


jshin47:
The CI is failing but I don't think it has anything to do with your code...

ronaldyuwandika:
I think the CI is failing because there's a problem with the docker build and push stage that needs username and password for dockerhub repository

deploy/charts/rawfile-csi/values.yaml
```diff
@@ -12,6 +12,10 @@ defaults: &defaults
     requests:
       cpu: 10m
       memory: 100Mi
+  rawfile_dir: /rawfile
```
SeMeKh, line 15: Please change this to `/var/csi/rawfile` to keep it backward-compatible.

Dockerfile
```diff
@@ -19,5 +19,7 @@ ENV PYTHONUNBUFFERED 1
 
 ARG IMAGE_TAG
 ARG IMAGE_REPOSITORY
+ARG DATA_DIR
```
SeMeKh, line 22: Let's not make data_dir configurable. This is what the container sees, and does not need to be changed.

SeMeKh (changes requested):
Thanks for your contribution. Will you please fix the comments?

deploy/charts/rawfile-csi/values.yaml
ronaldyuwandika: sure, I will change it later

Dockerfile
ronaldyuwandika: Ah okay, I'll undo this change

deploy/charts/rawfile-csi/values.yaml
ronaldyuwandika: hello, @SeMeKh I've updated this to `/var/csi/rawfile`
can you please check it again?
thank you

deploy/charts/rawfile-csi/templates/01-node-plugin.yaml
```diff
@@ -66,6 +66,8 @@ spec:
               value: unix:///csi/csi.sock
             - name: IMAGE_REPOSITORY
               value: "{{ .Values.node.image.repository }}"
+            - name: DATA_DIR
```
SeMeKh, line 69: The "remove data_dir variable" applies to this one too

deploy/charts/rawfile-csi/templates/01-node-plugin.yaml
```diff
@@ -85,7 +87,7 @@ spec:
               mountPath: /var/lib/kubelet
               mountPropagation: "Bidirectional"
             - name: data-dir
-              mountPath: /data
+              mountPath: {{ .Values.node.data_dir }}
```
SeMeKh, line 90: ditto

deploy/charts/rawfile-csi/values.yaml
```diff
@@ -12,6 +12,10 @@ defaults: &defaults
     requests:
       cpu: 10m
       memory: 100Mi
+  rawfile_dir: /var/csi/rawfile
+  data_dir: /data
```
SeMeKh, line 16: ditto

templates/task.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       imagePullPolicy: IfNotPresent
       volumeMounts:
         - name: data-dir
-          mountPath: /data
+          mountPath: {data_dir}
```
SeMeKh, line 23: ditto

SeMeKh (approved):
Thanks. Fix these and it's good to go.

templates/task.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   volumes:
     - name: data-dir
       hostPath:
-        path: /var/csi/rawfile
+        path: {rawfile_dir}
```
SeMeKh, line 14: Actually, this needs to be passed into the templating function so that it can get rendered.

SeMeKh (changes requested):

templates/task.yaml
ronaldyuwandika: image
can i change it into templating function like this?

SeMeKh: The values are interpolated not by helm, but the python code. So that's where the code needs to be updated.

PR #6, opened 2021-01-30
Fix minor issues related to resources

hanieh-m:
What this PR does:
- Adds `resources` to sidecar containers in chart in order to limit sidecar containers as well.
- Removes `requests` from task pods so that pods don't get stuck in pending state and pvcs are able to resize successfully.


SeMeKh (approved):
lgtm

PR #10, opened 2021-11-29
Allow the data directory to be customized

### LINKED ISSUE #3: Make the data-dir configurable.

taliesins:
Allow the data directory to be customized
Default location of install to a location where Docker won't run out of disk space
Modify helm chart to use hooks to control installation/uninstallation order when used with other charts (there can be issues when uninstalling charts otherwise)

This closes #3


mhkarimi1383:
@tiagolobocastro,

Can you review this?
 We have already covered data dir customization, but here we have other changes too, and I don't know if we need them or not since I have not seen such things in other OpenEBS Projects

tiagolobocastro:
There's lots of changes here in addition to "allow configurable data directory"
I have not seen them, but if it makes sense we can consider them having them as an optional setting.
I would suggest we close and reopen separate smaller PRs if @taliesins is still interested.

mhkarimi1383:
So, I will close this and we will kindly ask @taliesins to open smaller PRs for the changes since we have also refactored the helm chart recently

PR #13, opened 2022-06-29
add arm64 support

odidev:
The following file has been created and modified:

Added support for arm64 in .travis.yml to make docker images and publish them on docker hub.

**Note:** I have not added the e2e-testing for the arm64 platform because e2e-testing [requires minikube](https://github.com/odidev/rawfile-localpv/blob/0eb4eefd13aca1b3dd5a3e102ec43c1b91509393/.ci/e2e-test/setup.sh#L13) to run on travis for arm64 platform. And travisCI does not support minikube on the arm64 platform. There is an [open issue](https://github.com/kubernetes/minikube/issues/11845) for this in kubernetes/minikube.

Signed-off-by: odidev odidev@puresoftware.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 os: linux
+services:
+  - docker
+
 dist: bionic
 language: minimal
 jobs:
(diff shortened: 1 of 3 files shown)


odidev:
Hi Team,
A gentle reminder to take a look at the above changes.

.travis.yml
```diff
@@ -18,3 +21,21 @@ jobs:
...
+      env:
+        - CHANGE_MINIKUBE_NONE_USER=true
+        - MINIKUBE_WANTUPDATENOTIFICATION=false
+        - MINIKUBE_WANTREPORTERRORPROMPT=false
+        - MINIKUBE_HOME=$HOME
+        - CHANGE_MINIKUBE_NONE_USER=true
+        - KUBECONFIG=$HOME/.kube/config
+    - stage: publish-arm
+      arch: arm64
+      script: ./.ci/publish.sh
```
semekh, line 41: Looks like a code duplication to me. Do you think we could avoid that somehow?

semekh (changes requested):
Thanks for your contribution. Generally LGTM, but it'd be great if we could avoid the code duplication.

odidev:
> Looks like a code duplication to me. Do you think we could avoid that somehow? 

I think we need to go with this only because when I try to reduce the duplication it does not run for all the stages.

odidev:
Hi Team,
Adding a gentle reminder to take a look at the above changes.

mhkarimi1383:
@tiagolobocastro 
I think we can close this since we want to migrate to GitHub Actions (https://github.com/openebs/rawfile-localpv/issues/41)

tiagolobocastro:
> @tiagolobocastro I think we can close this since we want to migrate to GitHub Actions (#41)

Yes, I can add arm64 from the getgo to GitHub Actions.

PR #22, opened 2022-12-28
add mkfs nodiscard option to accelerate disk formatting phase

mchtech:
Some disk backends support `discard` feature to release unused blocks on underlying storage.
It will take a long time to discard blocks when using `mkfs` to format, if large "rawfile" is created on these storages.

This patch adds `-E nodiscard / -K` parameter to `mkfs.xxx` to accelerate "rawfile" formatting phase.

### DIFF
--- a/declarative.py
+++ b/declarative.py
@@ -70,9 +70,9 @@ def current_fs(device):
 def be_formatted(dev, fs):
     def init_fs(device):
         if fs == "ext4":
-            run(f"mkfs.ext4 -m 0 {device}")
+            run(f"mkfs.ext4 -m 0 -E nodiscard {device}")
         elif fs == "btrfs":
-            run(f"mkfs.btrfs {device}")
+            run(f"mkfs.btrfs -K {device}")
             tmp_mnt = tempfile.mkdtemp(prefix="mnt-")
             default_subvol = f"{tmp_mnt}/default"
             run(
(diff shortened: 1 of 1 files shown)


mhkarimi1383:
I think it's better to do discard, but you can make an option to provide `extraFormattingOptions` in the configuration or storageclass parameters

tiagolobocastro:
When is this an issue, if the disk itself doesn't support trim, and then loopback writes zeroes, is that what happens?

Yes, I also think this should be configurable.

mhkarimi1383:
> When is this an issue, if the disk itself doesn't support trim, and then loopback writes zeroes, is that what happens?
> 
> Yes, I also think this should be configurable.

That's why I think we should not hard code any non default OS parameters

mhkarimi1383:
@tiagolobocastro 
I want to add a parameter to enable or disable `nodiscard` option, IDK that it should be Global in `csi-driver` or in storage class parameters

tiagolobocastro:
> @tiagolobocastro I want to add a parameter to enable or disable `nodiscard` option, IDK that it should be Global in `csi-driver` or in storage class parameters

I would say should be storage class parameters, maybe something like the formatOptions which you had been working on for lvm?

mhkarimi1383:
> > @tiagolobocastro I want to add a parameter to enable or disable `nodiscard` option, IDK that it should be Global in `csi-driver` or in storage class parameters
> 
> I would say should be storage class parameters, maybe something like the formatOptions which you had been working on for lvm?

Yeah, but it was easier to handle that via global configuration

mhkarimi1383:
Closing this since we have planned to add support for formatOptions

PR #27, opened 2024-02-22
✨ feat: add stroage-class to helm chart

mhkarimi1383:
Adding StorageClass to Helm Chart will make this chart to be a great all in one setup for provisioning cluster storages

I also wanted to make helm chart docs to automatically generated using https://github.com/norwoodj/helm-docs
and also using GitHub Actions as a helm repo will remove the need of the cloning project to install its chart, If you are Okay with them I would love to work on them :)
Also I want to make the host path for data dir customizable


mhkarimi1383:
@semekh 
I'm Closing this PR and I will create a new one with all of them applied and using make it based of develop branch

PR #16, opened 2022-11-29
Fix xfs_grow arguments

laurivosandi:
Looks like device path is passed instead of the mountpoint.

PR #34, opened 2024-07-26
Fix xfs expansion bug

guoard:
The `xfs_growfs` utility requires the path of a mounted XFS filesystem, not the device path, to perform an expansion.

#### Previous Behavior

```bash
$ xfs_growfs -d device

xfs_growfs: device is not a mounted XFS filesystem
```


guoard:
closed in favour of: #16

PR #38, opened 2025-01-03
fix: update pb files after protobuf was upgraded

### LINKED ISSUE #37: csi-driver crashed for old pb files
```log
TypeError: Descriptors cannot not be created directly. If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0." It is followed by the proposition to install the appropriate version of protobuf library.

If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
```

mchtech:
fix: #37 

https://github.com/openebs/rawfile-localpv/commit/ff598bf4aa1b9e0f00a92a7ddbabfe801ec4d619 upgraded `protobuf`, so pb files should be regenerated.


Dockerfile
```diff
@@ -18,6 +18,7 @@ ADD ./requirements.txt ./
 RUN pip install -r ./requirements.txt
 
 ADD ./ ./
+RUN python3 /usr/local/lib/python3.11/site-packages/grpc_tools/protoc.py --proto_path=protos/ protos/csi.proto --grpc_python_out=csi/ --python_out=csi/
```
mhkarimi1383, line 21: I think it should not be there

When we change pb files we have to do this and push changes
(We need some CI Tests for checking of that)
also python version is hard-coded

tiagolobocastro (approved):

Dockerfile
tiagolobocastro: You mean we should generate the pb file outside, and COPY it inside?

mhkarimi1383: Yes,
Also it is pointing to site-packages directly, it's better to use python -m instead
I think I will refactor Dockerfile to use Docker multi stage more efficiently

mhkarimi1383: Or we have to separate the stage for the generation

tiagolobocastro: Yes, we also use python -m on [mayastor](https://github.com/openebs/mayastor-control-plane/blob/9a2b3a1e12cb09237c0003dfecc0f57b984fa28a/scripts/python/venv-setup-prep.sh#L105), but it's only for testing there, since the running code is in rust

mhkarimi1383: Yeah, but here it is not for test, and it is used for production Dockerfile

tiagolobocastro:
Cherry-picked to #62

PR #63, opened 2025-03-05
Moving from requirements.txt to pyproject.toml (poetry)

### LINKED ISSUE #45: Switch from requirements.txt to pyproject.toml

### LINKED ISSUE #46: Update Python dependencies

mhkarimi1383:
Closes https://github.com/openebs/rawfile-localpv/issues/45
Closes #46

Also replaced https://github.com/kdave/btrfs-progs/archive/refs/tags/v6.3.2.tar.gz#egg=btrfsutil&subdirectory=libbtrfsutil/python with native packages


mhkarimi1383:
Due to https://github.com/kdave/btrfs-progs/issues/964 we are unable to move `btrfsutil` from bare URL to be a standard dependency

mhkarimi1383:
Will squash them if needed ;)

Dockerfile
```diff
@@ -1,31 +1,56 @@
...
-ENV PYTHONUNBUFFERED=1
-
+COPY docker-entrypoint.sh /docker-entrypoint.sh
 
 ARG IMAGE_TAG
 ARG IMAGE_REPOSITORY
 ENV IMAGE_REPOSITORY=${IMAGE_REPOSITORY}
 ENV IMAGE_TAG=${IMAGE_TAG}
+
+ENTRYPOINT /docker-entrypoint.sh $0 $@
```
tiagolobocastro: can we run this with `tini` or something like that? I noticed that the pods aren't getting sigtermed gracefully

shell.nix
```diff
@@ -11,6 +11,20 @@ pkgs.mkShell {
...
     kind
+    python311
+    (poetry.override { python3 = python311; })
+    gcc
+    gnumake
+    btrfs-progs
+    stdenv.cc.cc.lib
   ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure") [ docker-client ];
+  shellHook = ''
+    export LD_LIBRARY_PATH="${pkgs.stdenv.cc.cc.lib}/lib"
```
tiagolobocastro, line 22: are there duplicate cc or is it like this?

rawfile/bd2fs.py
tiagolobocastro, line 1: 👍

tiagolobocastro (approved):
Nice one 🚀

shell.nix
mhkarimi1383: It was needed :)
(For btrfs progs and their libs)

Dockerfile
mhkarimi1383: I think it will get fixed by using the json array form of ENTRYPOINT

mhkarimi1383:
I have also upgraded packages too here

Dockerfile
mhkarimi1383: @tiagolobocastro 

It should be OK now, I have change entrypoint to use json array

mhkarimi1383:
CI is stuck because of crashes I think 🤔

tiagolobocastro:
> CI is stuck because of crashes I think 🤔

Seems like it, maybe we should add --fail-fast for PRs?
On failure you'll see a dump archive attached to the workflow with the logs to help you debug the failures

mhkarimi1383:
> > CI is stuck because of crashes I think 🤔
> 
> Seems like it, maybe we should add --fail-fast for PRs?
> On failure you'll see a dump archive attached to the workflow with the logs to help you debug the failures

Yeah they should fail faster or we should check and dump health of the pods before running tests

tiagolobocastro:
> > > CI is stuck because of crashes I think 🤔
> > 
> > 
> > Seems like it, maybe we should add --fail-fast for PRs?
> > On failure you'll see a dump archive attached to the workflow with the logs to help you debug the failures
> 
> Yeah they should fail faster or we should check and dump health of the pods before running tests

Some cli arg related failures it seems, you can check the dump now.

> we should check and dump health of the pods before running tests

We're only running the K8s e2e tests (which were the existing travis tests), so not sure if we can run anything custom. What we should do is add some of our own tests, maybe some simple sanity tests to start with, and if those fail no point running the other ones.

mhkarimi1383:
Yeah tests are running using some ready to use tools that are used for Kubernetes storage provisioner test, I will check them out...

I think here the problem is due to changes to packaging and its effect on relative imports ,etc.

I have to refactor module structure and imports...

mhkarimi1383:
Found the problem

I older runs pods where restarting before running test, And now this is not the case

mhkarimi1383:
@tiagolobocastro 

Almost done.

Will merge this if nothing left

PR #64, opened 2025-03-09
CI: Migrate from Travis to GitHub Actions

### LINKED ISSUE #41: Migrate Travis to GithubActions
CI testing is not currently enabled.
Looks like this repo used to have a [Travis CI](https://github.com/openebs/rawfile-localpv/blob/develop/.travis.yml) which is no longer relevant.
We should migrate this to GitHub Actions similar to other OpenEBS repos.

tiagolobocastro:
Migrates travis to github actions by adding workflow files for PR, push and scheduled runs.
todo: Helm Chart specific testing doesn't really exist and should be added.
(I've added some basic helm chart linting and template test)

Resolves #41


mhkarimi1383:
I think you can move the checkout step to ci.yaml file to run single time

tiagolobocastro:
> I think you can move the checkout step to ci.yaml file to run single time

Not quite sure how to do that, since they're separate jobs?

mhkarimi1383:
We can do it by using a workflow call trigger, but it will increase complexity

tiagolobocastro:
> We can do it by using a workflow call trigger, but it will increase complexity

Then I'd suggest let's leave it as is now, can always change it if need be.

.ci/common
```diff
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
 set -ex
```
mhkarimi1383, line 4: Adding a check for needed tools will make it awesome using 

```shell
command -v docker > /dev/null || ( echo 'docker not found. exiting...'; exit 1 )

```

tiagolobocastro:
@mhkarimi1383 anything else to do here?

mhkarimi1383:
LGTM :)

mhkarimi1383 (approved):

.ci/common
tiagolobocastro: Indeed, done

mhkarimi1383:
Running tests after a while 🎉

mhkarimi1383:
Also Helm chart publish is missing

tiagolobocastro:
> Also Helm chart publish is missing

Yes indeed, I'll take this up as a separate PR

PR #73, opened 2025-03-13
:arrow_up: Upgrade python in pyproject and nix shell

mhkarimi1383:


shell.nix
```diff
@@ -11,8 +11,8 @@ pkgs.mkShell {
     helm-docs
     nixos-shell
     kind
-    python311
-    (poetry.override { python3 = python311; })
+    python313
+    poetry # Python3.13 is not supported (Overriding python3 input will not work)
```
tiagolobocastro, line 15: if 3.13 is not supported how will this work after this change then? poetry uses a different version itself?

mhkarimi1383: Yeah, but it uses 3.13 for the project itself, it will get an error for mypy setup if I override python3 input

tiagolobocastro (approved):

PR #74, opened 2025-03-14
:sparkles: Add fsType to storageClass in helm

mhkarimi1383:


deploy/charts/rawfile-csi/templates/storageclass.yaml
```diff
@@ -9,6 +9,8 @@ provisioner: {{ $vals.provisionerName }}
 reclaimPolicy: {{ $class.reclaimPolicy }}
 volumeBindingMode: {{ $class.volumeBindingMode }}
 allowVolumeExpansion: {{ $class.allowVolumeExpansion }}
+parameters:
+  csi.storage.k8s.io/fstype: {{ $class.fsType | default "ext4" }}
```
tiagolobocastro, line 13: Is it worth validating support fs?

Something like this:
```
{{- $fs := $class.fsType | default "ext4" }}
{{- if (regexMatch "^((ext4|xfs|btrfs))$" $fs) -}}
```

deploy/charts/rawfile-csi/values.yaml
```diff
@@ -139,6 +139,8 @@ storageClasses:
     allowVolumeExpansion: true
     # -- Sets default reclaimPolicy for StorageClass volumes
     reclaimPolicy: Delete
+    # -- Sets filesystem type for volumes
```
tiagolobocastro: also document which are the supported ones?

deploy/charts/rawfile-csi/templates/storageclass.yaml
mhkarimi1383: I think we should handle that using json schema

tiagolobocastro: Yeah, that would be even better :)

mhkarimi1383: Will do it in another PR

tiagolobocastro (approved):

PR #79, opened 2025-04-09
Update to CSI spec v1.10.0

addyess:
### Overview
* Update the CSI proto from [v1.10.0](https://github.com/container-storage-interface/spec/blob/v1.10.0/csi.proto)
* Update the default sig-storages images used by the chart
* Updates the CI to test with 3 Kubernetes Versions based on availability from kind=0.27.0


addyess:
@tiagolobocastro i don't understand the failure here.  I can perform resizes with this image without issue.


Best i can understand is that we're using the e2e suite:

https://github.com/kubernetes/kubernetes/blob/v1.27.0/test/e2e/storage/testsuites/volume_expand.go#L218

my guess is that its waiting `pvcConditionSyncPeriod= 2 * time.Minute` for something to change on the PVC

that something specifically is the PVC's Status condition to have a type `PersistentVolumeClaimFileSystemResizePending`

tiagolobocastro:
There's a suspicious log in external resize:
`W0411 19:31:37.442375       1 warnings.go:70] unknown field "status.allocatedResourceStatuses"`

tiagolobocastro:
Maybe this?
https://github.com/kubernetes-csi/external-resizer/issues/491

addyess:
> Maybe this? [kubernetes-csi/external-resizer#491](https://github.com/kubernetes-csi/external-resizer/issues/491)

that's it precisely.  I see it now.  Maybe we can update the kubernetes we're testing on?  i see currently we're using `1.27.0`.  That seems pretty ancient now doesn't it as we approach 1.33. 

i see the test is fixed in [1.32.1](https://github.com/kubernetes/kubernetes/blob/v1.32.1/test/e2e/storage/testsuites/volume_expand.go)

I guess there's a few options:
* bump the tests to a more modern kubernetes
* downgrade from using the latest external-resizers

tiagolobocastro:
> > Maybe this? [kubernetes-csi/external-resizer#491](https://github.com/kubernetes-csi/external-resizer/issues/491)
> 
> that's it precisely. I see it now. Maybe we can update the kubernetes we're testing on? i see currently we're using `1.27.0`. That seems pretty ancient now doesn't it as we approach 1.33.
> 
> i see the test is fixed in [1.32.1](https://github.com/kubernetes/kubernetes/blob/v1.32.1/test/e2e/storage/testsuites/volume_expand.go)
> 
> I guess there's a few options:
> 
>     * bump the tests to a more modern kubernetes
> 
>     * downgrade from using the latest external-resizers

Looks like it was merged to 1.30, 1.31 and 1.32 (the 3 supported minor)
Probably we should keep a matrix of versions to test, but for now let's just bump it

.github/workflows/k8s.yml
```diff
@@ -5,6 +5,15 @@ on:
...
     runs-on: ubuntu-latest-8-cores
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: 
+        - v1.32.3
+        - v1.31.6
+        - v1.30.10
+    env:
+      K8S_VERSION: ${{ matrix.k8s_version }}
```
addyess: Testing with multiple kubernetes versions at the same time

nix/sources.json
```diff
@@ -1,14 +1,14 @@
...
-        "rev": "31571599597cf5bd035928576b6abe922451ce7d",
-        "sha256": "1m4b3bgbch8isxq01h6ghz61jrfwf7cvp13srxgsin7jxm6hrxiw",
+        "rev": "ca4a1e055eaf8051cfc5b299f3a01ec16f2fde58",
+        "sha256": "0pp6bhvbbh5mz7fhg7ajnin4ak6411ry78mrnmhykgyp9nkdhj0c",
         "type": "tarball",
-        "url": "https://github.com/NixOS/nixpkgs/archive/31571599597cf5bd035928576b6abe922451ce7d.tar.gz",
+        "url": "https://github.com/NixOS/nixpkgs/archive/ca4a1e055eaf8051cfc5b299f3a01ec16f2fde58.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     }
 }
```
addyess: this was the first commit/branch i could find that supported a kind recent enough (0.27.0) to include kubernetes versions that have the volume expansion test fixed.  I understand if this is too "bleeding edge" to pull in.  But if the tests pass -- at least the commit and sha256 are correct

mhkarimi1383:
Yeah, I agree to bump up K8s version to use one of the last three versions that are supported since I have already upgraded dependencies, I will make a PR and update testing tools and K8s version that is used in tests

addyess:
@mhkarimi1383 are you saying you'll take the changes here i've made to the testing and integrate into a separate PR from this one?  Or can this PR be merged as is?

mhkarimi1383:
@addyess 
I think it's better to do upgrades in a separated PR and then you can update your branch

addyess:
@mhkarimi1383 no worries. need me to raise a separate PR?  It sounds like you wanted to handle it.  Either way.  Good to see the tests pass.  

🚀

mhkarimi1383:
@addyess I will take care of this :)

nix/sources.json
mhkarimi1383: @addyess See #82 for that We will take care of that there

mhkarimi1383: We will not use master branch (We can use unstable at least)

.github/workflows/k8s.yml
mhkarimi1383: I have moved this into a file in #82 
inspired by some others like .node-version or .python-version

But it's good to test against last three versions

tiagolobocastro (approved):

mhkarimi1383 (approved):

PR #82, opened 2025-04-12
:wrench: Making K8s version configurable

mhkarimi1383:
And we have better documentation of what version we are testing this project against

also I have used one version left to the end to be both update and compatible (https://kubernetes.io/releases/)


mhkarimi1383:
CC: @addyess

mhkarimi1383:
NixPKGs upgrade is needed since we need a newer version of kind (https://github.com/kubernetes-sigs/kind/issues/3871)


@tiagolobocastro For that I have to move to the unsable branch of the nixpkgs or downloading kind manually in shell hook or I can add another source named unstable and use that for kind

tiagolobocastro:
> NixPKGs upgrade is needed since we need a newer version of kind ([kubernetes-sigs/kind#3871](https://github.com/kubernetes-sigs/kind/issues/3871))
> 
> @tiagolobocastro For that I have to move to the unsable branch of the nixpkgs or downloading kind manually in shell hook or I can add another source named unstable and use that for kind

Does K8s v1.30.10 work with the current kind version?

mhkarimi1383:
@tiagolobocastro 
No, because they have added 1.29 and later on last release that they have upgraded the containerd to v2

https://github.com/kubernetes-sigs/kind/releases/tag/v0.27.0

mhkarimi1383:
We can use versions that are listed here

https://github.com/kubernetes-sigs/kind/releases/tag/v0.24.0

tiagolobocastro:
We can add an unstable channel just for kind.
You can add a channel with niv to the nix/sources.json and then pull kind from there.

mhkarimi1383:
@tiagolobocastro 

Problem solved, But the LD_PATH that we are setting here it's making some other programs to not function correctly when versions are different from the global packages

tiagolobocastro:
You mean `LD_LIBRARY_PATH`?

Looks like CI passed at least, which programs are not working correctly?

mhkarimi1383:
Yes

`div` itself
`gitmoji-cli` (Because of NodeJS)
And LSPs that are enabled in NeoVim

for example when I want to run `nvim` I have to run `LD_LIBRARY_PATH="" nvim`

We should use nix-ld I think

tiagolobocastro:
Doesn't seem to change for me, whether I add the kind from unstable or not, library path is always the same.

mhkarimi1383:
> Doesn't seem to change for me, whether I add the kind from unstable or not, library path is always the same.


I use master branch of nixpkgs in local machine (https://github.com/mhkarimi1383/nixos)

So I see version conflict of gcc and shared libs

tiagolobocastro:
> > Doesn't seem to change for me, whether I add the kind from unstable or not, library path is always the same.
> 
> I use master branch of nixpkgs in local machine (https://github.com/mhkarimi1383/nixos)
> 
> So I see version conflict of gcc and shared libs

And you see that only after making this change? or do you see it even in develop?

tiagolobocastro (approved):

mhkarimi1383:
Yes, I see that after and before this change on every branch :(

mhkarimi1383:
@addyess,
Can you also review this since this change will affect your changes I think

addyess (approved):
I sure don't know much about nix packages -- but the outcome i see is a single k8s version i need to test my updates. 

👍🏻

PR #84, opened 2025-04-17
:art: Add pre-commit and format the code

mhkarimi1383:
For CI to work we need to install pre-commit github app


.github/workflows/fossa.yml
```diff
@@ -15,4 +15,3 @@ jobs:
       - uses: fossas/fossa-action@v1.4.0
         with:
           api-key: ${{ secrets.FOSSA_API_KEY }}
-
```
tiagolobocastro, line 18: any reason to remove all the newlines at the end? not having them can cause some "issues" on git diff

tiagolobocastro (approved):

.github/workflows/fossa.yml
mhkarimi1383: It is an popular hook that removes EOF, And I have added this since I have used it a lot

mhkarimi1383: See https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#end-of-file-fixer and https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/end_of_file_fixer.py

tiagolobocastro: Yeah I see, I'd prefer to keep the eof to avoid unnecessary diff when modifying last lines of a file but if you want to remove them then it's also fine

mhkarimi1383: I saw that we have extra new line on some files, and not having same rule in every file is not a great idea I think

tiagolobocastro: Yes it'd be good to be consistent throughout.
If we keep this validation then let's also include the rawfile/csi/ as well.

mhkarimi1383: I have excluded them since they where some generated files

tiagolobocastro: could we exclude those ones generated only?
anyway, this is fine lgtm

mhkarimi1383: Yeah, I has support for globing also

mhkarimi1383:
@tiagolobocastro
Can you take care of installing pre-commit app for this repo, with that we will have a ci for that

tiagolobocastro:
> @tiagolobocastro Can you take care of installing pre-commit app for this repo, with that we will have a ci for that

Done

PR #86, opened 2025-04-21
:recycle: Refactor FS Support and Utilities to be more standard

### LINKED ISSUE #54: Using interfaces for file system support
Currently for things like snapshot, extend, etc. we are having functions per file system, We can make this standard by creating an abstract interface to implement them more cleanly

mhkarimi1383:
Closes #54


mhkarimi1383:
Hi

@addyess 
@tiagolobocastro 

I have made a draft base class for making filesystem support extendable

mhkarimi1383:
Some duplicate codes are here and old codes will be removed

rawfile/filesystem/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+        return data["filesystems"][0]["target"]
+    except subprocess.CalledProcessError:
+        return None
+
+
+def from_device(device: str) -> FilesystemBase:
+    res = subprocess.run(
+        f"blkid -o value -s TYPE {device}", shell=True, capture_output=True
+    )
+    if res.returncode == 2:  # specified token was not found
```
tiagolobocastro: What does this mean, specified token was not found?

rawfile/filesystem/base.py
```diff
@@ -0,0 +1,82 @@
...
+            "unmount",
+            "resize",
+        ]
+        for method in required_methods:
+            if not (hasattr(subclass, method) and callable(getattr(subclass, method))):
+                return False
+        return True
+
+    @abstractmethod
+    def create_snapshot(self):
```
tiagolobocastro: Should these take the snapshot uuid?

rawfile/filesystem/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+filesystems.update(
+    {
+        "ext4": EXT4,
+        "btrfs": BTRFS,
+        "xfs": XFS,
+    }
+)
+
+
+def device_to_mountpoint(device: str) -> None | str:
```
tiagolobocastro, line 27: Is it fine to always return the first mountpoint? Ex: does it matter if it's the globalmount or the per-pod bind mount?

rawfile/filesystem/base.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Create a snapshot of the current state of the file system.
+        """
+        raise NotSupportedError("create_snapshot")
+
+    @abstractmethod
+    def delete_snapshot(self):
+        """
+        Delete the most recent snapshot of the file system.
```
tiagolobocastro: Should this be the snapshot specified rather than the most recent?

rawfile/filesystem/base.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        raise NotSupportedError("format_fs")
+
+    @abstractmethod
+    def mount(self, mountpoint: str, options: list[str] = []):
+        """
+        Mount the file system.
+
+        Should also do
+            self.mountpoint = mountpoint
```
tiagolobocastro, line 196: but here the mountpoint in args != self.mountpoint right?

rawfile/filesystem/base.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @abstractmethod
+    def delete_snapshot(self):
+        """
+        Delete the most recent snapshot of the file system.
+        """
+        raise NotSupportedError("delete_snapshot")
+
+    @abstractmethod
+    def format_fs(self, options: list[str] = []):
```
tiagolobocastro: What would format_fs do if we already have the filesystem mountpoint?
Ah because it can be None? So we should only allow format_fs is mountpoint is None

tiagolobocastro (commented):
this is good, we had also added some filesystem interface to mayastor

rawfile/filesystem/__init__.py
mhkarimi1383: Yeah it is find since we don't want to touch data, We just want to manage the life 
cycle

rawfile/filesystem/__init__.py
mhkarimi1383: It means that device does not exists

rawfile/filesystem/base.py
mhkarimi1383: I have just wrote the Draft to see what functions do we need since I have not read the snapshotting part completely

rawfile/filesystem/base.py
mhkarimi1383: Yeah, It should have an input for id/name

rawfile/filesystem/base.py
mhkarimi1383: Yeah, We should format before mount like what we have in mount utils in go

rawfile/filesystem/base.py
mhkarimi1383: Yeah, Also self.mountpoint could be None

rawfile/filesystem/base.py
mhkarimi1383: format and mount has been added

mhkarimi1383:
@tiagolobocastro 
After adding block support CI is not passing IDK why, But it was passing in it's PR

I have added `ext4` filesystem class

tiagolobocastro:
> @tiagolobocastro After adding block support CI is not passing IDK why, But it was passing in it's PR
> 
> I have added `ext4` filesystem class

Yeah, seems like there's no loop devices for the volume_path during the node extend... doesn't always happen and I'm not sure how that can even happen!
I'll try to run the test on the loop and see if I can work out how this is happening...

tiagolobocastro:
> @tiagolobocastro After adding block support CI is not passing IDK why, But it was passing in it's PR
> 
> I have added `ext4` filesystem class

One bug I hit, not sure if it's the one on  CI, but it looks like the RO flag on a loop device is sticky sometimes, for some reason, so we probably need to ensure the loop is rw after creating it.
Also I figured out how to handle RO on publish properly, we need to create another RO loop device, rather than a using the bind mount... will check if this is doable or not..

mhkarimi1383:
> figured

Nice, I was thinking about the loop device limit but I saw that default limit is `2^20` on the kernel (I think) https://elixir.bootlin.com/linux/v6.14.6/source/include/linux/kdev_t.h#L7

if user don't set the max_loop parameter

mhkarimi1383:
And I see the problem happens randomly :)

tiagolobocastro:
> And I see the problem happens randomly :)

Oh why is it?
I'm running on a loop for a few days again, but no luck this time, at least on my system :(

mhkarimi1383:
> > And I see the problem happens randomly :)
> 
> Oh why is it?
> I'm running on a loop for a few days again, but no luck this time, at least on my system :(

I think the resize test error happens randomly
And I think it is when the order of test runs changes.

mhkarimi1383:
@tiagolobocastro
Hi I have added All of the filesystem, Can you review this?

Then I will connect other parts with new classes

mhkarimi1383:
Also for btrfs we had some options that are added by default just like the previous implication

It should be removed when we have this option?

mhkarimi1383:
After this, I want to Switch to official `kubernetes` Python lib

mhkarimi1383:
@tiagolobocastro 
The resize problem happened again :(

mhkarimi1383:
I think it's because of the test run order/concurrency

In some states we are unable to resize I think

tiagolobocastro:
> @tiagolobocastro The resize problem happened again :(

I'm trying to repro on branch ci-fix, but doesn't seem to happen

mhkarimi1383:
> > @tiagolobocastro The resize problem happened again :(
> 
> I'm trying to repro on branch ci-fix, but doesn't seem to happen

That's weird, I think we have some sort of raise condition or something like that.

Not for this part, but we could have raise condition for the metadata files that we have in the nodes (In other provisioners we are using CRDs)

I will make metadata files to have lock files or I will switch t something like SQLite

mhkarimi1383:
I Going to connect other parts with this part of the code if everything is OK

mhkarimi1383:
@tiagolobocastro 

Looks like the Resize problem is not resolved :(

We can create an issue for tracking this I think

tiagolobocastro:
> I Going to connect other parts with this part of the code if everything is OK

Seems reasonable, though there's a lot of changes on the btrfs snapshot. We should probably enable e2e tests for snapshots before making the changes.

mhkarimi1383:
> > I Going to connect other parts with this part of the code if everything is OK
> 
> Seems reasonable, though there's a lot of changes on the btrfs snapshot. We should probably enable e2e tests for snapshots before making the changes.

Yeah, I have moved from raw commands to it's library, I have to also check how the snapshot restore works and also how to configure that to only test snapshot on supported filesystems

tiagolobocastro:
> @tiagolobocastro
> 
> Looks like the Resize problem is not resolved :(
> 
> We can create an issue for tracking this I think

I'm still unable to reproduce it locally.
But I've spotted a few bugs related to loop devices, which may explain it, I've raised a PR.

mhkarimi1383:
> > I Going to connect other parts with this part of the code if everything is OK
> 
> Seems reasonable, though there's a lot of changes on the btrfs snapshot. We should probably enable e2e tests for snapshots before making the changes.

As for testing btrfs snapshotting, I have not seen any option to test snapshotting (specially for a single FS) in https://github.com/kubernetes/kubernetes/blob/v1.16.0/test/e2e/storage/testsuites/testdriver.go#L139-L185 and https://github.com/kubernetes/kubernetes/blob/v1.16.0/test/e2e/storage/external/external.go#L142-L211

I'm looking to find a way to test this feature

mhkarimi1383:
Or we can use multiple test configs for each driver, Or we should implement our own tests

Also we can submit an issue for Kubernetes :)

mhkarimi1383:
@tiagolobocastro

We should upgrade nixpkgs, As you can see everytime we get updates from dependabot lock file changes, we should run `poetry sync` in our local setups

mhkarimi1383:
Getting

```shell
ImportError: /opt/pysetup/.venv/lib/python3.13/site-packages/btrfsutil.cpython-313-x86_64-linux-gnu.so: undefined symbol: btrfs_util_subvolume_is_valid
```

in controller pod

tiagolobocastro:
> Getting
> 
> ```shell
> ImportError: /opt/pysetup/.venv/lib/python3.13/site-packages/btrfsutil.cpython-313-x86_64-linux-gnu.so: undefined symbol: btrfs_util_subvolume_is_valid
> ```
> 
> in controller pod

Seems like some version incompatibility?

mhkarimi1383:
@tiagolobocastro I have opened an issue on btrfs-progs, I will Check what is the problem, But the last solution will be using bare commands

mhkarimi1383:
@tiagolobocastro I have opened an issue on btrfs-progs, I will Check what is the problem, But the last soloution will be using bare commands

mhkarimi1383:
Since my IDE does not bother me about importing stuff

mhkarimi1383:
I have checked that the installed clib is correctly exporting the required thing

I think I have to switch to commands for that since resolving this problem will take much time

tiagolobocastro:
I can reproduce it too, it seems
Also btw, the LD_PRELOAD removal seems to not work for me:
```
❯ cd rawfile; ./rawfile.py                                                                                                                                                                                                                                                                                               ─╯
Traceback (most recent call last):
  File "/home/tiago/git/openebs/rawfile-localpv/rawfile/./rawfile.py", line 5, in <module>
    import bd2fs
  File "/home/tiago/git/openebs/rawfile-localpv/rawfile/bd2fs.py", line 3, in <module>
    import grpc
  File "/home/tiago/.cache/pypoetry/virtualenvs/rawfile-xNvq-e8_-py3.13/lib/python3.13/site-packages/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/home/tiago/.cache/pypoetry/virtualenvs/rawfile-xNvq-e8_-py3.13/lib/python3.13/site-packages/grpc/_compression.py", line 20, in <module>
    from grpc._cython import cygrpc
ImportError: libstdc++.so.6: cannot open shared object file: No such file or directory
```

mhkarimi1383:
The error message looks different in your reproduce logs, will checkout if this is happening for me or not

tiagolobocastro:
btw is `btrfs_util_subvolume_is_valid` the correct method?

that seems to be the C api where in the python api we have `is_subvolume` which does seem to work:

```
 File "/home/tiago/git/openebs/rawfile-localpv/rawfile/./rawfile.py", line 64, in <module>
    if btrfsutil.is_subvolume("/volume"):
       ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
btrfsutil.BtrfsUtilError: [BtrfsUtilError 11 Errno 2] Could not statfs: No such file or directory: '/volume'
```

mhkarimi1383:
```
strings /opt/pysetup/.venv/lib/python3.13/site-packages/btrfsutil.cpython-313-x86_64-linux-gnu.so  | grep btrfs_util_subvolume_is_valid 
btrfs_util_subvolume_is_valid
btrfs_util_subvolume_is_valid_fd
btrfs_util_subvolume_is_valid_fd
btrfs_util_subvolume_is_valid
btrfs_util_subvolume_is_valid
btrfs_util_subvolume_is_valid_fd
```

mhkarimi1383:
Nice thing is that from now, Filesystem operation errors are more readable :)

```

Traceback (most recent call last):
  File "/opt/pysetup/.venv/lib/python3.13/site-packages/grpc/_server.py", line 610, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/rawfile/util.py", line 41, in wrap
    raise exc
  File "/rawfile/util.py", line 22, in wrap
    res = func(self, request, context)
  File "/rawfile/bd2fs.py", line 103, in NodeStageVolume
    fs.format_and_mount(
    ~~~~~~~~~~~~~~~~~~~^
        mount_options=[]
        ^^^^^^^^^^^^^^^^
    )  # TODO: Respect from bd_publish_request.volume_capability
    ^
  File "/rawfile/filesystem/base.py", line 174, in format_and_mount
    self.mount(mountpoint, mount_options)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/rawfile/filesystem/ext4.py", line 53, in mount
    raise FileSystemMountError(self.__filesystem__, e.stderr.decode())
filesystem.base.FileSystemMountError: Operation 'mount' failed on filesystem 'ext4': mount: /var/lib/kubelet/plugins/kubernetes.io/csi/rawfile.csi.openebs.io/042646f75e48f744fc998ed869378b5a50589802da6b6ef5a4ad782b21533829/globalmount/mount: /dev/loop0 already mounted on /var/lib/kubelet/plugins/kubernetes.io/csi/rawfile.csi.openebs.io/042646f75e48f744fc998ed869378b5a50589802da6b6ef5a4ad782b21533829/globalmount/mount.
       dmesg(1) may have more information after failed mount system call.
```

mhkarimi1383:
@tiagolobocastro 
I'm adding a verifier for already mounted volumes

Is it something to be in the filesystem package or is it needed for other parts? (Like block volumes)

mhkarimi1383:
I have created a utility to get the device by the mountpoint

mhkarimi1383:
@tiagolobocastro The artifact name of the cluster dump has conflict the name of cluster dump of the smoke test is the same as the e2e test

tiagolobocastro:
> @tiagolobocastro The artifact name of the cluster dump has conflict the name of cluster dump of the smoke test is the same as the e2e test

ah yes, fixing on the other PR: https://github.com/openebs/rawfile-localpv/pull/128/commits/1eb8823e927b6e8a23a620d107f8437ff4483378

tiagolobocastro:
I added the snapshot test, but the snapshot was not being picked up. Turns out we're missing the snapshot controller, so I added it along with rbac changes for snapshotcontent and now I hit the issue: 
`failed to take snapshot of the volum │
│ e pvc-4a92b1cc-670c-4171-bc62-61fe5b6f437b: "rpc error: code = Unknown desc = Exception calling application: /opt/pysetup/.venv/lib/python3.13/site-packages/btrfsutil.cpython │
│ -313-x86_64-linux-gnu.so: undefined symbol: btrfs_util_subvolume_is_valid"`

tiagolobocastro:
Aha, I got it working somehow

mhkarimi1383:
> I added the snapshot test, but the snapshot was not being picked up. Turns out we're missing the snapshot controller, so I added it along with rbac changes for snapshotcontent and now I hit the issue: 
> `failed to take snapshot of the volum │
> │ e pvc-4a92b1cc-670c-4171-bc62-61fe5b6f437b: "rpc error: code = Unknown desc = Exception calling application: /opt/pysetup/.venv/lib/python3.13/site-packages/btrfsutil.cpython │
> │ -313-x86_64-linux-gnu.so: undefined symbol: btrfs_util_subvolume_is_valid"`

Please add them and comment them out :)

I will uncomment them after my refactor on the snapshot :)

mhkarimi1383:
@tiagolobocastro 


Something happened that is so strange

```

ERROR:grpc._server:Exception calling application: Operation 'format' failed on filesystem 'xfs': unknown option -m 0
Usage: mkfs.xfs
/* blocksize */		[-b size=num]
/* config file */	[-c options=xxx]
/* metadata */		[-m crc=0|1,finobt=0|1,uuid=xxx,rmapbt=0|1,reflink=0|1,
			    inobtcount=0|1,bigtime=0|1]
/* data subvol */	[-d agcount=n,agsize=n,file,name=xxx,size=num,
			    (sunit=value,swidth=value|su=num,sw=num|noalign),
			    sectsize=num
/* force overwrite */	[-f]
/* inode size */	[-i perblock=n|size=num,maxpct=n,attr=0|1|2,
			    projid32bit=0|1,sparse=0|1,nrext64=0|1]
/* no discard */	[-K]
/* log subvol */	[-l agnum=n,internal,size=num,logdev=xxx,version=n
			    sunit=value|su=num,sectsize=num,lazy-count=0|1]
/* label */		[-L label (maximum 12 characters)]
/* naming */		[-n size=num,version=2|ci,ftype=0|1]
/* no-op info only */	[-N]
/* prototype file */	[-p fname]
/* quiet */		[-q]
/* realtime subvol */	[-r extsize=num,size=num,rtdev=xxx]
/* sectorsize */	[-s size=num]
/* version */		[-V]
			devicename
<devicename> is required unless -d name=xxx is given.
<num> is xxx (bytes), xxxs (sectors), xxxb (fs blocks), xxxk (xxx KiB),
      xxxm (xxx MiB), xxxg (xxx GiB), xxxt (xxx TiB) or xxxp (xxx PiB).
<value> is xxx (512 byte blocks).
Traceback (most recent call last):
  File "/rawfile/filesystem/xfs.py", line 20, in format_fs
    output = subprocess.run(
        f"mkfs.{self.__filesystem__} {' '.join(options)} {self.device}",
    ...<2 lines>...
        capture_output=True,
    )
  File "/usr/local/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'mkfs.xfs -m0 /var/lib/kubelet/plugins/kubernetes.io/csi/rawfile.csi.openebs.io/abefcfaa14320b22b5d75c6dab14c9956218d97f04aa7ac8c0e2fccea43d1c7f/globalmount/device' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/pysetup/.venv/lib/python3.13/site-packages/grpc/_server.py", line 610, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/rawfile/util.py", line 41, in wrap
    raise exc
  File "/rawfile/util.py", line 22, in wrap
    res = func(self, request, context)
  File "/rawfile/bd2fs.py", line 103, in NodeStageVolume
    fs.format_and_mount(
    ~~~~~~~~~~~~~~~~~~~^
        mount_options=[]
        ^^^^^^^^^^^^^^^^
    )  # TODO: Respect from bd_publish_request.volume_capability
    ^
  File "/rawfile/filesystem/base.py", line 175, in format_and_mount
    self.format_fs(format_options)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/rawfile/filesystem/xfs.py", line 28, in format_fs
    raise FileSystemFormatError(self.__filesystem__, e.stderr.decode())
filesystem.base.FileSystemFormatError: Operation 'format' failed on filesystem 'xfs': unknown option -m 0
Usage: mkfs.xfs
/* blocksize */		[-b size=num]
/* config file */	[-c options=xxx]
/* metadata */		[-m crc=0|1,finobt=0|1,uuid=xxx,rmapbt=0|1,reflink=0|1,
			    inobtcount=0|1,bigtime=0|1]
/* data subvol */	[-d agcount=n,agsize=n,file,name=xxx,size=num,
			    (sunit=value,swidth=value|su=num,sw=num|noalign),
			    sectsize=num
/* force overwrite */	[-f]
/* inode size */	[-i perblock=n|size=num,maxpct=n,attr=0|1|2,
			    projid32bit=0|1,sparse=0|1,nrext64=0|1]
/* no discard */	[-K]
/* log subvol */	[-l agnum=n,internal,size=num,logdev=xxx,version=n
			    sunit=value|su=num,sectsize=num,lazy-count=0|1]
/* label */		[-L label (maximum 12 characters)]
/* naming */		[-n size=num,version=2|ci,ftype=0|1]
/* no-op info only */	[-N]
/* prototype file */	[-p fname]
/* quiet */		[-q]
/* realtime subvol */	[-r extsize=num,size=num,rtdev=xxx]
/* sectorsize */	[-s size=num]
/* version */		[-V]
			devicename
<devicename> is required unless -d name=xxx is given.
<num> is xxx (bytes), xxxs (sectors), xxxb (fs blocks), xxxk (xxx KiB),
      xxxm (xxx MiB), xxxg (xxx GiB), xxxt (xxx TiB) or xxxp (xxx PiB).
<value> is xxx (512 byte blocks).


```


Found in logs, But this option is only set for `ext4`

mhkarimi1383:
Found the problem, I had to do deepcopy :')

The format and mount part is connected successfully

mhkarimi1383:
Resize part is now connected too. and I'm removing the functions that we are not using them after refactor

mhkarimi1383:
@tiagolobocastro 
For mount and unmount since the logic is shared between the filesystem and block volumes, I will add a mounter for block volumes separately and decide which one to use in the handler

mhkarimi1383:
@tiagolobocastro
I think we are having storage problem in GH Actions, Having the insufficient storage error here and in other PR's Pipelines

mhkarimi1383:
Also Don't know why we have `controllerExpansion` enabled

tiagolobocastro:
> @tiagolobocastro I think we are having storage problem in GH Actions, Having the insufficient storage error here and in other PR's Pipelines

hmm I wonder if it's because I removed a pre-setup step which I thought was not required, but I guess it must be then! I'll bring it back shortly

tiagolobocastro:
> Also Don't know why we have `controllerExpansion` enabled

We do indeed do controller expansion via CSI if that's what you mean

mhkarimi1383:
> > @tiagolobocastro I think we are having storage problem in GH Actions, Having the insufficient storage error here and in other PR's Pipelines
> 
> hmm I wonder if it's because I removed a pre-setup step which I thought was not required, but I guess it must be then! I'll bring it back shortly

I should debug it further, Since it happened after my last commit, But We have same problem in other PRs too.

mhkarimi1383:
> > Also Don't know why we have `controllerExpansion` enabled
> 
> We do indeed do controller expansion via CSI if that's what you mean

We have both `controllerExpansion` and `nodeExpansion` and when running expansion from controller, We need to use task pods

tiagolobocastro:
> > @tiagolobocastro I think we are having storage problem in GH Actions, Having the insufficient storage error here and in other PR's Pipelines
> 
> hmm I wonder if it's because I removed a pre-setup step which I thought was not required, but I guess it must be then! I'll bring it back shortly

Seems I was mistaken, was never added on rawfile... I'll try upgrading the action workers then

tiagolobocastro:
> > > Also Don't know why we have `controllerExpansion` enabled
> > 
> > 
> > We do indeed do controller expansion via CSI if that's what you mean
> 
> We have both `controllerExpansion` and `nodeExpansion` and when running expansion from controller, We need to use task pods

yes because node expansion is only triggered when pod is published
we could probably replace the task pod with grpc calls

mhkarimi1383:
> > > > Also Don't know why we have `controllerExpansion` enabled
> > > 
> > > 
> > > We do indeed do controller expansion via CSI if that's what you mean
> > 
> > 
> > We have both `controllerExpansion` and `nodeExpansion` and when running expansion from controller, We need to use task pods
> 
> yes because node expansion is only triggered when pod is published we could probably replace the task pod with grpc calls

Also we can make commands of task pods to run modules/functions directly using `python -m ...`

mhkarimi1383:
My problem solved, It was a bug in new logic for unmount

But we still have storage problems in other PRs and IDK why

tiagolobocastro:
> My problem solved, It was a bug in new logic for unmount
> 
> But we still have storage problems in other PRs and IDK why

The enterprise runner fails to install stuff, not sure why..
So I'm just trying to reproduce the failure with ssh enabled so I can check why we're running out of space, but it doesn't want to fail now :)

mhkarimi1383:
@tiagolobocastro 
We have a failure already there https://github.com/openebs/rawfile-localpv/pull/138

tiagolobocastro:
> @tiagolobocastro We have a failure already there #138

Yeah but there's no ssh at the end so I can debug the state

mhkarimi1383:
> > @tiagolobocastro We have a failure already there #138
> 
> Yeah but there's no ssh at the end so I can debug the state

Oh, I See :)

mhkarimi1383:
```
ERROR:grpc._server:Exception calling application: Operation 'mount' failed on filesystem 'ext4': mount: /var/lib/kubelet/pods/8157044d-6523-4c33-8622-ef0566a1f865/volumes/kubernetes.io~csi/pvc-81b5694a-81b8-4fa1-a469-6a878e26120f/mount: /dev/loop1 already mounted on /var/lib/kubelet/plugins/kubernetes.io/csi/rawfile.csi.openebs.io/28d3beb10310814218d9e0e1b42d6fea91c1856954a7a04d93a87bf76b00f7b5/globalmount/mount.
       dmesg(1) may have more information after failed mount system call.
Traceback (most recent call last):
  File "/rawfile/filesystem/ext4.py", line 54, in mount
    output = subprocess.run(
        f"mount -t {self.__filesystem__} {' '.join(options)} {self.device} {mountpoint}",
    ...<2 lines>...
        capture_output=True,
    )
  File "/usr/local/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'mount -t ext4 -o ro /dev/loop1 /var/lib/kubelet/pods/8157044d-6523-4c33-8622-ef0566a1f865/volumes/kubernetes.io~csi/pvc-81b5694a-81b8-4fa1-a469-6a878e26120f/mount' returned non-zero exit status 32.
```

ro mounts for filesystems where ignored in the older mount implementation, We should ignore them for now too?

mhkarimi1383:
After mount problem fixes the only left part will be snapshots :)

tiagolobocastro:
> ```
> ERROR:grpc._server:Exception calling application: Operation 'mount' failed on filesystem 'ext4': mount: /var/lib/kubelet/pods/8157044d-6523-4c33-8622-ef0566a1f865/volumes/kubernetes.io~csi/pvc-81b5694a-81b8-4fa1-a469-6a878e26120f/mount: /dev/loop1 already mounted on /var/lib/kubelet/plugins/kubernetes.io/csi/rawfile.csi.openebs.io/28d3beb10310814218d9e0e1b42d6fea91c1856954a7a04d93a87bf76b00f7b5/globalmount/mount.
>        dmesg(1) may have more information after failed mount system call.
> Traceback (most recent call last):
>   File "/rawfile/filesystem/ext4.py", line 54, in mount
>     output = subprocess.run(
>         f"mount -t {self.__filesystem__} {' '.join(options)} {self.device} {mountpoint}",
>     ...<2 lines>...
>         capture_output=True,
>     )
>   File "/usr/local/lib/python3.13/subprocess.py", line 577, in run
>     raise CalledProcessError(retcode, process.args,
>                              output=stdout, stderr=stderr)
> subprocess.CalledProcessError: Command 'mount -t ext4 -o ro /dev/loop1 /var/lib/kubelet/pods/8157044d-6523-4c33-8622-ef0566a1f865/volumes/kubernetes.io~csi/pvc-81b5694a-81b8-4fa1-a469-6a878e26120f/mount' returned non-zero exit status 32.
> ```
> 
> ro mounts for filesystems where ignored in the older mount implementation, We should ignore them for now too?

Up to you, can leave it for another PR as well
Still need to check if we can impl RO for block as well, though that one seems to be a lot more complicated

mhkarimi1383:
I have tried RO mounts multiple times but it was a failure, We should work on that in another PR

rawfile/bd2fs.py
```diff
@@ -249,19 +250,29 @@ def ControllerExpandVolume(self, request, context):
...
-        volume_id = request.source_volume_id
-        name = request.name
-
-        snapshot_id, creation_time_ns = btrfs_create_snapshot(
-            volume_id=volume_id, name=name
-        )
+        fs = None
+        try:
+            file = img_file(request.source_volume_id)
+            loop_dev = attach_loop(file)
```
tiagolobocastro, line 256: we should also detach the source file after taking the snapshot?

rawfile/bd2fs.py
```diff
@@ -271,7 +282,19 @@ def CreateSnapshot(self, request: csi_pb2.CreateSnapshotRequest, context):
 
     @log_grpc_request
     def DeleteSnapshot(self, request: csi_pb2.DeleteSnapshotRequest, context):
+        fs = None
         snapshot_id = request.snapshot_id
         volume_id, name = snapshot_id.rsplit("/", 1)
-        btrfs_delete_snapshot(volume_id=volume_id, name=name)
+        try:
+            file = img_file(volume_id)
+            loop_dev = attach_loop(file)
```
tiagolobocastro, line 291: same question as above on create

rawfile/rawfile_util.py
```diff
@@ -39,9 +38,9 @@ def img_file(volume_id):
 def destroy(volume_id, dry_run=True):
     print(f"Destroying {volume_id}")
     if not dry_run:
-        be_absent(img_file(volume_id))
-        be_absent(meta_file(volume_id))
-        be_absent(img_dir(volume_id))
+        Path(img_file(volume_id)).unlink()
```
tiagolobocastro: is unlink idempotent? ie will it fail if it's already removed?

rawfile/declarative.py
```diff
@@ -31,39 +32,58 @@ def be_symlink(path, to):
...
-        if True:  # TODO: verify that the right device is mounted
-            return
-        # noinspection PyUnreachableCode
-        be_unmounted(mountpoint)
-    elif mountpoint.resolve().is_file():
+    if current_dev:
+        current_dev = Path(current_dev).resolve().as_posix()
+        if current_dev == device:
+            return  # TODO: LOG
+        unmount(mountpoint, clear_mountpoint=False)
```
tiagolobocastro: I wonder if here we should abort instead? if something else is using the same mountpath could be dangerous to proceed

rawfile/filesystem/btrfs.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            with TemporaryDirectory(
+                prefix=f"rawfile.{self.__filesystem__}.fs-snapshot."
+            ) as tmp_dir:
+                try:
+                    snapshots_dir = f"{tmp_dir}/.snapshots"
+                    snapshot_subvol = f"{snapshots_dir}/{name}"
+                    subprocess.run(
+                        f"""
+                        set -ex
```
tiagolobocastro: Probably safer to set pipefail?

```suggestion
                        set -exo pipefail
```

rawfile/filesystem/btrfs.py
```diff
@@ -0,0 +1,180 @@
...
+            return output.stdout.decode()
+        except subprocess.CalledProcessError as e:
+            raise FileSystemMountError(self.__filesystem__, e.stderr.decode())
+        except Exception as e:
+            raise FileSystemMountError(self.__filesystem__, str(e))
+
+    def unmount(self, clear_mountpoint=False) -> str:
+        try:
+            if get_device_for_mountpoint(self.mountpoint) != self.device:
+                return f"{self.device} is not mounted at {self.mountpoint}"
```
tiagolobocastro: if `clear_mountpoint` then we should also rmdir the mountpoint here for idempotency

deploy/helm/rawfile-localpv/README.md
```diff
@@ -33,6 +33,7 @@ Kubernetes: `>= 1.21`
 | controller.tolerations | list | `[{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Equal","value":"true"}]` | Tolerations for controller component |
 | crds.csi.volumeSnapshots.enabled | bool | `true` | Install Volume Snapshot CRDs |
 | crds.enabled | bool | `true` | Disables the installation of all CRDs if set to false |
+| defaultFileSystem | string | `"ext4"` | Used when `fsType` is not set in the StorageClass |
```
tiagolobocastro: how was this being handled "today"?

rawfile/declarative.py
mhkarimi1383: You mean we should raise some error?
The TODO said that We should unmount :)

rawfile/filesystem/btrfs.py
mhkarimi1383: Already done at rawfile/filesystem/btrfs.py:96

rawfile/rawfile_util.py
mhkarimi1383: It will fail, I will now fix it

rawfile/bd2fs.py
mhkarimi1383: We should, but I have not found that it was done in older implantation. will fix that

deploy/helm/rawfile-localpv/README.md
mhkarimi1383: It will get stored in `default_fs` key in config (`rawfile/rawfile.py:32`), then it will be used as fallback here `rawfile/filesystem/__init__.py`

mhkarimi1383:
@tiagolobocastro 

I think `ReadWriteOncePod` is already covered in tests, Why it is not checked in README yet?

Also Offline expansion is working, but filesystem resize will apply after first use of volume, and it is natural for now I think

deploy/helm/rawfile-localpv/README.md
mhkarimi1383: It is hardcoded to ext4 now
In the get_fs function that I have removed

rawfile/declarative.py
mhkarimi1383: Or we can call `fsync {mountpoint}`, then unmount

But this should not happen :)

mhkarimi1383:
@tiagolobocastro 
Got the storage error in Smoke Tests while bootstrapping cluster ;)

rawfile/bd2fs.py
```diff
@@ -249,19 +251,36 @@ def ControllerExpandVolume(self, request, context):
...
+                )
+
+            fs.create_snapshot(name=request.name)
+            creation_time_ns = time.time_ns()
+            snapshot_id = f"{request.source_volume_id}/{request.name}"
+        finally:
+            if fs:
+                fs.unmount(clear_mountpoint=True)
+            if loop_dev:
+                subprocess.run(
```
tiagolobocastro: I think I just realized we may not be able to do this, what if the volume was actually mounted already?

rawfile/declarative.py
tiagolobocastro: Yeah I'd suggest raising some error and let someone handle it, should it ever happen

rawfile/declarative.py
```diff
@@ -31,39 +32,58 @@ def be_symlink(path, to):
...
+    #     options.extend(["-o", "ro"])
+    fs.mount(options=options)
+
+
+def unmount(mountpoint: str, clear_mountpoint=True):
+    path = Path(mountpoint)
+    if not path.exists():
+        return
+    if path.is_file() or path.is_block_device():
+        subprocess.run(
```
tiagolobocastro: should we use the utils "run" methods as used elsewhere in the code?
would allow us to add trace logging in a common place for verbose logging

rawfile/filesystem/btrfs.py
tiagolobocastro: I mean if `get_device_for_mountpoint` returns `None`
This can happen in umount call gets stuck for a long time and csi-node pod is deleted

rawfile/filesystem/xfs.py
```diff
@@ -0,0 +1,93 @@
...
+            raise FileSystemFormatError(self.__filesystem__, e.stderr.decode())
+        except Exception as e:
+            raise FileSystemFormatError(self.__filesystem__, str(e))
+
+    def mount(
+        self,
+        mountpoint: str | None = None,
+        options: list[str] = [],
+    ) -> str:
+        if mountpoint is None:
```
tiagolobocastro: there's a lot of code that seems duplicated, I wonder if we could have a default implementation that we could reuse as xfs/ext4 code would be mostly the same

tiagolobocastro (approved):
The commit list is pretty huge (52), could you tidy up or squash them?

tiagolobocastro:
> @tiagolobocastro Got the storage error in Smoke Tests while bootstrapping cluster ;)

hmm:
```
/dev/sdb1        74G   71G     0 100% /mnt
```
Not sure why it's so full, I wonder if we could clean up the /mnt?

mhkarimi1383:
> The commit list is pretty huge (52), could you tidy up or squash them?

Will squash that

rawfile/bd2fs.py
mhkarimi1383: We are using a whole new loop block dev here, It should be Ok

rawfile/declarative.py
mhkarimi1383: I will raise some error

rawfile/declarative.py
mhkarimi1383: That was due to a circular import problem, I will create a whole new utils module, I will categorize utils into multiple submodules, And I will use a global command runner

rawfile/filesystem/btrfs.py
mhkarimi1383: Oh, yeah it would not remove that when we get into this situation, will move that to finally so it will run anytime

rawfile/filesystem/xfs.py
mhkarimi1383: Yeah, will fix that too!

rawfile/bd2fs.py
tiagolobocastro: Doesn't the existing loop dev get reused?

mhkarimi1383: No we are attaching a new loop dev

tiagolobocastro: hmm I thought if device already exists it's reused for idempotency

rawfile/declarative.py
mhkarimi1383: @tiagolobocastro
The utilities are refactored now, and I have moved all of the `run` calls to use a single function

rawfile/bd2fs.py
mhkarimi1383: No, it is a new independent device

rawfile/filesystem/xfs.py
mhkarimi1383: Exception cast has been changed, we are having less duplicates now, I will remove more duplicates...

mhkarimi1383: Shared logics are moved into the base class

tiagolobocastro (approved):

mhkarimi1383:
@tiagolobocastro 

If everything is OK, I will squash my commits and merge

rawfile/bd2fs.py
tiagolobocastro: This is checking already existing attached devices first?

```python
            devs = attached_loops(file)
            if len(devs) > 0:
                # we could use -L to ensure there's no overlap, and thus having
                # only 1 device at most a match and allowing us to simply use
                # losetup --direct-io=on -fL --show {file}
                dev = devs[0]
                # sometimes a RO attribute is sticky on the loop device, for some reason
                run(f"blockdev --setrw {dev}")
                return dev
```

mhkarimi1383:
@tiagolobocastro 

Squashed :)

tiagolobocastro (approved):

tiagolobocastro:
@mergifyio queue

PR #92, opened 2025-05-10
build: use shell python for poetry

tiagolobocastro:
Since we are using a different python version than the one which comes packaged with poetry, we ensure it's set.


mhkarimi1383:
Thanks, forgot to do that, I used to run it manually

mhkarimi1383 (approved):

PR #93, opened 2025-05-10
feat(mode/block): add block mode support

tiagolobocastro:
Since the rawfile's building block is a loop device, adding block mode support is relatively straightforward.

The node publish creates the target_path as a file, rather than a directory
The be_mounted, does a bind mount rather than a filesystem one The node stage ofc skips the formatting step
The volume stats collects only the device size
The node expand stops after loop device expansion

### DIFF
--- a/rawfile/bd2fs.py
+++ b/rawfile/bd2fs.py
@@ -17,7 +17,8 @@ from declarative import (
     be_mounted,
     be_unmounted,
 )
-from fs_util import mountpoint_to_dev, path_stats
+from fs_util import AccessType, mountpoint_to_dev, path_stats
+from rawfile_util import attached_loops
 from google.protobuf.timestamp_pb2 import Timestamp
 from remote import btrfs_create_snapshot, btrfs_delete_snapshot
 from util import log_grpc_request
@@ -28,6 +29,18 @@ def get_fs(request):
     if fs_type == "":
         fs_type = "ext4"
     return fs_type
+
+
+def get_access_type(request):
+    access_type = request.volume_capability.WhichOneof("access_type")
+    return check_access_type(access_type)
+
+
+def check_access_type(access_type):
+    try:
+        return AccessType[access_type]
+    except KeyError:
+        raise Exception(f"Unsupported access type: {access_type}")
 
 
 class Bd2FsIdentityServicer(csi_pb2_grpc.IdentityServicer):
@@ -58,8 +71,16 @@ class Bd2FsNodeServicer(csi_pb2_grpc.NodeServicer):
     @log_grpc_request
     def NodePublishVolume(self, request, context):
         staging_dev = f"{request.staging_target_path}/device"
-        Path(request.target_path).mkdir(exist_ok=True)
-        be_mounted(dev=staging_dev, mountpoint=request.target_path)
+
+        path = Path(request.target_path)
+        access_type_actions = {
+            AccessType.mount: path.mkdir,
+            AccessType.block: path.touch,
+        }
+        access_type_actions[get_access_type(request)](exist_ok=True)
+        be_mounted(
+            dev=staging_dev, mountpoint=request.target_path, readonly=request.readonly
+        )
         return csi_pb2.NodePublishVolumeResponse()
 
     @log_grpc_request
@@ -94,10 +115,11 @@ class Bd2FsNodeServicer(csi_pb2_grpc.NodeServicer):
 
         self.bds.NodePublishVolume(bd_publish_request, context)
 
-        mount_path = f"{request.staging_target_path}/mount"
-        Path(mount_path).mkdir(exist_ok=True)
-        be_formatted(dev=bd_publish_request.target_path, fs=get_fs(request))
-        be_mounted(dev=bd_publish_request.target_path, mountpoint=mount_path)
+        if get_access_type(request) is AccessType.mount:
+            mount_path = f"{request.staging_target_path}/mount"
+            Path(mount_path).mkdir(exist_ok=True)
+            be_formatted(dev=bd_publish_request.target_path, fs=get_fs(request))
+            be_mounted(dev=bd_publish_request.target_path, mountpoint=mount_path)
 
         return csi_pb2.NodeStageVolumeResponse()
 
@@ -125,6 +147,8 @@ class Bd2FsNodeServicer(csi_pb2_grpc.NodeServicer):
     # @log_grpc_request
     def NodeGetVolumeStats(self, request, context):
         volume_path = request.volume_path
+        if Path(volume_path).is_block_device():
+            return self.bds.NodeGetVolumeStats(request, context)
         stats = path_stats(volume_path)
         return csi_pb2.NodeGetVolumeStatsResponse(
             usage=[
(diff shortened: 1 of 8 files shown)


rawfile/bd2fs.py
```diff
@@ -122,9 +136,11 @@ def NodeUnstageVolume(self, request, context):
 
         return csi_pb2.NodeUnstageVolumeResponse()
 
-    # @log_grpc_request
+    @log_grpc_request
```
mhkarimi1383: Enabling this will produce alot of logs, We should imporove logging system later on.

tiagolobocastro: oops indeed this was for testing only, forgot to revert

rawfile/declarative.py
```diff
@@ -40,6 +41,14 @@ def be_mounted(dev, mountpoint):
             return
         # noinspection PyUnreachableCode
         be_unmounted(mountpoint)
+    elif mountpoint.resolve().is_file():
+        opts = ["bind"]
+        # don't we need to add this?
+        # if readonly:
```
tiagolobocastro: the mount mode also doesn't set ro, is this somehow handled by the kubelet or is this a bug?

tiagolobocastro:
Tests passing now, PTAL @mhkarimi1383

rawfile/util.py
```diff
@@ -36,10 +36,12 @@ def wrap(self, request, context):
 
 
 def run(cmd):
+    print(f"{cmd}")
```
mhkarimi1383: Please remove this, or make to have a better format

rawfile/util.py
```diff
@@ -36,10 +36,12 @@ def wrap(self, request, context):
...
     return subprocess.run(cmd, shell=True, check=True)
 
 
 def run_out(cmd: str):
+    print(f"{cmd}")
```
mhkarimi1383: Like the previous print :)

rawfile/declarative.py
mhkarimi1383: You mean that when we mount a block volume in readonly mode we will not receive that the volume mount is readonly?

rawfile/bd2fs.py
```diff
@@ -58,8 +67,14 @@ def NodeGetCapabilities(self, request, context):
     @log_grpc_request
     def NodePublishVolume(self, request, context):
         staging_dev = f"{request.staging_target_path}/device"
-        Path(request.target_path).mkdir(exist_ok=True)
-        be_mounted(dev=staging_dev, mountpoint=request.target_path)
+
+        if get_access_type(request) == "mount":
```
mhkarimi1383: Something like this will be readable later on, I think

```python
path = Path(request.target_path)
access_type_actions = {
  "mount": path.mkdir
  "block": path.touch
}
access_type_actions[get_access_type(request)](exist_ok=True)
```

rawfile/bd2fs.py
```diff
@@ -94,10 +109,11 @@ def NodeStageVolume(self, request, context):
 
         self.bds.NodePublishVolume(bd_publish_request, context)
 
-        mount_path = f"{request.staging_target_path}/mount"
-        Path(mount_path).mkdir(exist_ok=True)
-        be_formatted(dev=bd_publish_request.target_path, fs=get_fs(request))
-        be_mounted(dev=bd_publish_request.target_path, mountpoint=mount_path)
+        if get_access_type(request) == "mount":
```
mhkarimi1383: This one could be like the previous one :)

rawfile/bd2fs.py
```diff
@@ -145,6 +163,13 @@ def NodeGetVolumeStats(self, request, context):
 
     @log_grpc_request
     def NodeExpandVolume(self, request, context):
+        if get_access_type(request) == "block":
```
mhkarimi1383: Also this one

rawfile/bd2fs.py
```diff
@@ -204,7 +229,7 @@ def CreateVolume(self, request, context):
             )
 
         access_type = volume_capability.WhichOneof("access_type")
-        assert access_type == "mount"
+        assert access_type == "mount" or access_type == "block"
```
mhkarimi1383: We can use

```python
access_type in {"block", "mount"}
```

mhkarimi1383:
I made some reviews (Sorry, I hate If statement), Also we could have some global mappings/classes like what I'm doing for Filesystems, it will make it easier to develop and addore features later on

rawfile/bd2fs.py
mhkarimi1383: Or we should move to Interfaces

rawfile/util.py
tiagolobocastro: Can you suggest the format, currently without these logs it's very difficult to debug
Also there's no timestamps in the current logs, though this print will not address that :)

rawfile/bd2fs.py
tiagolobocastro: Done

rawfile/bd2fs.py
tiagolobocastro: How do you mean? Here it's just one if

rawfile/bd2fs.py
tiagolobocastro: Ah yes, forgot you can do that in python :)
Done

rawfile/bd2fs.py
mhkarimi1383: Yeah I was wrong

rawfile/declarative.py
tiagolobocastro: I mean, I don't see any code today for handling read only mount volumes, and there's no test failing, not sure why.
Either the tests aren't covering it, or it's a bug

tiagolobocastro: I just tested this, I set the readOnly of the PVC in a pod, and I see the readonly being set to true, but the device is still writeable...
Also though this commented out code here doesn't work, seems that the readonly flag is only in effect for the rawfile pod, and even then it's still writeable :)
Btw, added an enum for the access type, ptal

tiagolobocastro: Looks like we can't set the readonly flag on bind mounts for block devices, need to change the block device itself.
We can do this but... this wouldn't work when multiple pods have readonly flag set to a mix of true and false

rawfile/util.py
mhkarimi1383: We need a better logging system :)
IDK What we should do for now

rawfile/bd2fs.py
mhkarimi1383: Can you fix this one also :)

rawfile/util.py
tiagolobocastro: I'm not sure how these things are done in python, but there must be some sort of logging library which is tweakeable to some extent and which would include verbosity levels and timestamps?
I'll remove these for now, but we should think about adding something like that

mhkarimi1383: Yeah after filesystems I will change the logging system...

rawfile/declarative.py
mhkarimi1383: OK, I See, We should mention that we are not supporting readonly mounts

mhkarimi1383 (approved):

mhkarimi1383:
@tiagolobocastro 
If your done with this we can merge this :)

tiagolobocastro:
> @tiagolobocastro If your done with this we can merge this :)

thanks @mhkarimi1383, merged

PR #102, opened 2025-05-22
docs: add note for block volumes with RO

tiagolobocastro:
In order to support this with loop devices, the mount itself does not seem to be enough.
We likely need to create an additional loop device with RO set.


tiagolobocastro:
Alright, after the fix this is now skipping the CI @mhkarimi1383 :)

mhkarimi1383 (approved):

PR #106, opened 2025-05-26
A few bugfixes

tiagolobocastro:
fix(resize): resize img file first

    Ensures idempotency in case of crash between metadata change
    and file truncate.

---

    fix: remove mistaken -r and set rw on loop

    Remove a -r left in the mount by mistaken (not that it makes
    any difference, since the loop's RW/RO is what dictates it).
    Ensure the created loop device RW/RO is correct as I've seen
    that sometimes it becomes sticky.


mhkarimi1383 (approved):

mhkarimi1383:
Nice 😁

mhkarimi1383:
@tiagolobocastro 
We can merge this if nothing is left here ;)

PR #118, opened 2025-06-11
:technologist: Use nix-ld for `LD_LIBRARY_PATH`

mhkarimi1383:


mhkarimi1383:
@tiagolobocastro 
Here Can you also check it local development is affected in this way or not?

But user should enable nix-ld in it's global nix configuration

tiagolobocastro:
> @tiagolobocastro Here Can you also check it local development is affected in this way or not?
> 
> But user should enable nix-ld in it's global nix configuration

Affected in what way?
Why is it necessary to add nix-ld, won't that pick up things from the environment rather than the nix-shell?

mhkarimi1383:
> > @tiagolobocastro Here Can you also check it local development is affected in this way or not?
> > But user should enable nix-ld in it's global nix configuration
> 
> Affected in what way? Why is it necessary to add nix-ld, won't that pick up things from the environment rather than the nix-shell?

I mean the CLib(s) that we used are working correctly [Editor autocomplete, etc.] (like `btrfs`) 

Settings LD_LIBRARY_PATH will break some tools like `git`, `gitmoji` (`Node.js`)
Since it does not append value it just override the entire variable

tiagolobocastro:
hmm I see
Could we append the LD_LIBRARY_PATH instead of overriding it fully?

mhkarimi1383:
> hmm I see Could we append the LD_LIBRARY_PATH instead of overriding it fully?

No, Because LD_LIBRARY_PATH is empty by default, That's why I'm using `nix-ld`

shell.nix
```diff
@@ -2,10 +2,15 @@ let
   sources = import ./nix/sources.nix;
   pkgs = import sources.nixpkgs {};
   unstable-pkgs = import sources.nixpkgs-unstable {};
+  inherit (pkgs) lib stdenv;
 in
 pkgs.mkShell {
   name = "rawfile-shell";
 
+  NIX_LD_LIBRARY_PATH = lib.makeLibraryPath [
```
tiagolobocastro, line 10: Is this a special env name for nix?

tiagolobocastro:
> > hmm I see Could we append the LD_LIBRARY_PATH instead of overriding it fully?
> 
> No, Because LD_LIBRARY_PATH is empty by default, That's why I'm using `nix-ld`

Ah got it, because we're setting it at all is what's messing with other tools

shell.nix
mhkarimi1383: Yes, Checkout https://github.com/nix-community/nix-ld/blob/main/examples/masterpdfeditor.nix as a full example

tiagolobocastro (approved):

PR #120, opened 2025-06-13
:rocket: Helm Chart publish using Github Pages and actions

tiagolobocastro:
Adds new GHA release workflow:

A new release.ylm workflow orchestrates the release process into a number of steps:

1. Push Tag
1.a. check tag and version matches
2. Check it's a new release
3. Build and Push tagged image
4. K8s test using the tagged image
5. If something fails, then we should delete tag, add fixes and retag
6. Publish release and chart to gh-pages
7. Sanity check that the published chart is working

The build has also been modified for different registries and namespaces, meaning you can test this workflow from a fork. You only need to setup your own dockerhub credentials on the fork, or point it to ghcr.io


tiagolobocastro:
@mhkarimi1383 is this a good time to rename the chart from rawfile-csi into rawfile-localpv to match the package name? Also would match the other openebs engines.

Any thoughts @addyess ?

mhkarimi1383:
I completely agree  with this change :)

shell.nix
```diff
@@ -28,7 +29,9 @@ pkgs.mkShell {
     poetry env use $(which python)
     poetry install
     source $(poetry env info -p)/bin/activate
-    pre-commit install
+    if [ "${CI:-}" = "1" ]; then
```
mhkarimi1383: Is this reverse or something?

mhkarimi1383 (changes requested):

tiagolobocastro:
> oops sorry, pushed a tag by mistake, already deleted it

shell.nix
tiagolobocastro: doh, yeah that's wrong

tiagolobocastro: Done

tiagolobocastro:
@mhkarimi1383 PTAL, I think this is ready

mhkarimi1383:
@tiagolobocastro 
Nice :)
Approved 💯

mhkarimi1383 (approved):

PR #128, opened 2025-06-17
test: increase kubelet sync frequency

tiagolobocastro:
We've identified that the reason why the expansion tests take longer than expected is the kubelet sync frequency, which is what specifies how often the container and config are synced.

A high sync frequency could impact performance, but these are test clusters it's probably not an issue.

### DIFF
--- a/.ci/deployer.sh
+++ b/.ci/deployer.sh
@@ -113,6 +113,11 @@ $SUDO rm -rf "$TMP_KIND"/*
 cat <<EOF > "$TMP_KIND_CONFIG"
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+kubeadmConfigPatches:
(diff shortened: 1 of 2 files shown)


mhkarimi1383 (approved):

mhkarimi1383:
But E2E test duration has not changed if you compare it with previous PRs

tiagolobocastro:
> But E2E test duration has not changed if you compare it with previous PRs

Yeah I did not expect that, it's strange. I wonder if it's because it's already concurrent thus "bypassing" some of latency?
Also makes me wonder, we run 2 ginkgo commands today, could we merge it as 1? or perhaps even do it in separate github workflows

mhkarimi1383:
Also I think that ginkgo is able to run tests in parallel

PR #132, opened 2025-06-18
Helm: Add missing components for btrfs 

tiagolobocastro:
ci: add smoke tests for btrfs snapshots

---

    feat(helm): add missing snapshot controller

    The snapshot controller watches the K8s api for VolumeSnapshot and
    VolumeSnapshotContent CRD objects. It creates the contents which
    triggerds the csi snapshotter to create the snapshot.

---

    build(btrfsutil): downgrade to 6.13

    There's some issue with 6.14 causing:
    undefined symbol: btrfs_util_subvolume_is_valid


niladrih (approved):
LGTM

mhkarimi1383 (approved):

pyproject.toml
```diff
@@ -15,7 +15,7 @@ pyyaml = "^6.0.2"
 pykube-ng = "^23.6.0"
 munch = "^4.0.0"
 prometheus-client = "^0.22.1"
-btrfsutil = "^6.14"
+btrfsutil = "==6.13"
```
mhkarimi1383, line 18: Will remove it since maintaining this with these libs that are C based are complicated I think

I'm not using it in my newer filesystem support part

tiagolobocastro: Sure, I'll merge as is so the tests run for now, then you can remove it from your PR

PR #142, opened 2025-06-29
:technologist: Improve `vim.nix`

mhkarimi1383:
Disable k3s local storage
Use `.kube-version` file for k3s version
inherit `TERMINFO`


mhkarimi1383:
for example I'm using kitty, and it requires the terminfo to work, and since it is specific to my system, it is better to make it to inherit from main shell

tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

PR #143, opened 2025-06-29
:arrow_up: Upgrade k8s version to latest

mhkarimi1383:
We should be one version behind latest, but k3s_1_32 is not available


tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

PR #144, opened 2025-06-29
ci: use storage from root (sda1)

tiagolobocastro:
Sometimes /mnt (sdb1) is 100% used, which is failing the tests since we use /mnt for the test data.
Looks like GitHub docs seems to report only 14GiB of usable data, so I guess we can't rely on this sdb1.


mhkarimi1383:
@tiagolobocastro can you also checkout https://github.com/easimon/maximize-build-space/blob/master/action.yml?

PR #149, opened 2025-07-05
fix(snapshots): delete snapshot if source is deleted

tiagolobocastro:
If the source volume is not found, then we complete the delete snapshot. Also update bdd test to ensure we test deleting the snapshot. We're lacking a lot more snapshot testing...


mhkarimi1383 (approved):

tiagolobocastro:
@mergifyio queue

PR #150, opened 2025-07-05
Snapshot fixes

tiagolobocastro:
fix(init): ensure img-file has the correct size

    Fixes idempotency of volume creation in case file was created but
    not truncated to the size.

---

    test(smoke/snapshot): add snapshot deletion of unstaged volume

    Adds a test case where a volume is unstaged when its snapshot is deleted.

---

    fix(snapshot): idempotent snapshot delete

    On a retry the subvolume may already have been deleted.

---

    fix(snapshots): don't unmount volume on snap deletion

### DIFF
--- a/rawfile/bd2fs.py
+++ b/rawfile/bd2fs.py
@@ -21,7 +21,7 @@ from utils.logs import log_grpc_request
 from filesystem import get_from_device_or_fallback, from_device
 from filesystem.utils import get_device_for_mountpoint
 from rawfile_servicer import check_access_type, get_access_type
-from utils.rawfile import img_file, attach_loop, AccessType, path_stats
+from utils.rawfile import img_file, attach_loop, AccessType, path_stats, detach_loops
 from filesystem.base import UnknownFileSystemError
 from utils.lock import VolLock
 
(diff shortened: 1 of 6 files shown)


rawfile/filesystem/btrfs.py
```diff
@@ -103,7 +103,9 @@ def delete_snapshot(self, name: str):
                         default_subvol_id=$(btrfs subvolume get-default "{tmp_dir}" | awk '{{print $2}}')
                         default_subvol_path=$(btrfs subvolume list -o "{tmp_dir}" | awk -v id=$default_subvol_id '$2 == id {{print $NF}}')
                         default_subvol="{tmp_dir}/$default_subvol_path"
-                        btrfs subvolume delete "{snapshot_subvol}"
+                        if [ -f "{snapshot_subvol}" ]; then
```
tiagolobocastro, line 106: @mhkarimi1383 the vars created above don't seem to be "used", is it for debugging?

mhkarimi1383: I think I have forgotten to check them out after converting ops to use commands

rawfile/bd2fs.py
```diff
@@ -251,22 +251,18 @@ def ControllerExpandVolume(self, request, context):
...
-                raise UnknownFileSystemError(
-                    device=loop_dev, volume_id=request.source_volume_id
-                )
-
-            fs.create_snapshot(name=request.name)
-            creation_time_ns = time.time_ns()
-            snapshot_id = f"{request.source_volume_id}/{request.name}"
-        finally:
-            if fs:
-                fs.unmount(clear_mountpoint=True)
```
mhkarimi1383, line 286: Why we do not unmount anymore?

tiagolobocastro: there's nothing to unmount here, the mount which was being caught was from the volume or none if the volume is not staged in the node

rawfile/filesystem/btrfs.py
tiagolobocastro: Also I wonder if we should not use temporary directory, and just place one in well known location, ex: data folder
Otherwise if let's say the prod crashes during this operation, it'd leak a mount making it very difficult to get rid of.

rawfile/bd2fs.py
mhkarimi1383: Nice 👍

mhkarimi1383 (approved):

rawfile/filesystem/btrfs.py
mhkarimi1383: I think for these stuff we should do some race tests for example

tiagolobocastro:
@mergifyio queue

mhkarimi1383 (approved):

Issue #107, opened 2025-05-26
volume operation guard


tiagolobocastro:
If CSI calls get retried but previous operation is still ongoing, the new one should be rejected.
For this we should add per-volume locks, same as we have on other engines

mhkarimi1383:
Since currently we are not having CRDs, we should use `flock` or something like that.

tiagolobocastro:
Resolved by #151

PR #151, opened 2025-07-05
feat(lock): serialize volume operations

tiagolobocastro:
Ensures there's only 1 volume operation at a time. In case CSI times out, it may retry an operation, and thus we should ensure that there's no more than 1 at a time.
Also will help ensure the snapshot can safely determine when to remove a loop device.

### DIFF
--- a/rawfile/utils/lock.py
+++ b/rawfile/utils/lock.py
@@ -1,0 +1,26 @@
+import fcntl
+from utils.rawfile import lock_file
+
+"""
+  This can be used to ensure there's only 1 operation running at a time for a given volume.
+"""
+
+
+class VolLock:
+    def __init__(self, volume_id):
+        self.path = lock_file(volume_id)
+        self.path.touch(exist_ok=True)
+        self.file = open(self.path, "a")
+
+    def __enter__(self):
+        # grabs the lock or fails right away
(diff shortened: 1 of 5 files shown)


tiagolobocastro:
Resolves #107

rawfile/utils/lock.py
```diff
@@ -0,0 +1,25 @@
...
+from utils.rawfile import meta_file
+
+"""
+  This can be used to ensure there's only 1 operation running at a time for a given volume.
+"""
+
+
+class VolLock:
+    def __init__(self, volume_id):
+        self.path = meta_file(volume_id)
```
mhkarimi1383: What about using a separated `.lock` file

tiagolobocastro: What would the benefit of this be?

mhkarimi1383: Maybe it will not interfere with other ops on the same file, also it will be easier to clear lock file manually

tiagolobocastro: That's a good point, I've added a lock file, PTAL

mhkarimi1383 (approved):

rawfile/utils/lock.py
mhkarimi1383: Nice 😎

tiagolobocastro:
@mergifyio queue

PR #155, opened 2025-07-07
:technologist: Improve nix shell to fix shared libs problem

mhkarimi1383:
Wrapping python was not working because we are in a `venv`

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -6,10 +6,6 @@ in
 pkgs.mkShell {
   name = "rawfile-shell";
 
-  NIX_LD_LIBRARY_PATH = lib.makeLibraryPath [
-    pkgs.stdenv.cc.cc
-  ];
-
   buildInputs = with pkgs; [
     kubectl
     kubernetes-helm-wrapped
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -24,16 +20,19 @@ pkgs.mkShell {
...
+    export PYTHONPATH="$(git rev-parse --show-toplevel)/rawfile:$PYTHONPATH"
     source $(poetry env info -p)/bin/activate
     if ! [ "$CI" == "1" ]; then
       pre-commit install
     fi
   '';
   postShellHook = ''
     deactivate
+    unset PYTHONPATH
+    unset LD_PRELOAD
```
tiagolobocastro, line 36: Is this necessary?

tiagolobocastro (approved):

shell.nix
mhkarimi1383: I have added these to be sure that we are cleaning up :)

mhkarimi1383:
@Mergifyio queue

PR #156, opened 2025-07-07
:sparkles: Add reserved storage/capacity override support on nodes and thick provisioning and make it default

### LINKED ISSUE #9: Add option for available storage
With 0.7.0 release I saw there is tracking of available space and therefore no scheduling on a node with insufficient capacity happens (`RESOURCE_EXHAUSTED_EXIT_CODE`). This is great :)

Currently we have the following issue - it may happen that we have e.g. 100GB of storage on a node, single disk but we're need to define how much space we can use for rawdisks (e.g. 50%) because there are other components which may need that amount of storage defined as "reserved".

Would it be possible to provide this in future?

### LINKED ISSUE #24: Option to disable free space checks
With thin provisioning it's quite easy to bump into resource exhausted errors. 

I ended up disabling some checks

https://git.k-space.ee/k-space/rawfile-localpv/commit/17f653b15ff1edd43473f8c4f3adba33c3703d1a
https://git.k-space.ee/k-space/rawfile-localpv/commit/9c4cda7c0a33b7e1d65cda67608a09f6fe1fae36

Please make this configurable upstream

mhkarimi1383:
Resolves #9
Resolves #24


mhkarimi1383:
@tiagolobocastro 

For testing this, I think we should add a test case where we patch it and set reserved storage to `100%` and it should not create volumes

rawfile/utils/units.py
```diff
@@ -0,0 +1,27 @@
...
+    Converts a human-readable size string (e.g., "2GB", "500MB", "1.5KB", "5GiB")
+    to its equivalent value in bytes.
+    """
+    pretty_size = pretty_size.strip().upper()
+    if not pretty_size:
+        return 0
+
+    units = {
+        "B": 1,
+        "KB": 1024,
```
tiagolobocastro: Units without `i` should be base 10 not 2.
We can add the `i` variants as well

rawfile/utils/units.py
```diff
@@ -0,0 +1,27 @@
...
+        "MB": 1024**2,
+        "GB": 1024**3,
+        "TB": 1024**4,
+        "PB": 1024**5,
+    }
+    for unit, multiplier in reversed(units.items()):
+        for suffix in (unit, unit.replace("B", "IB")):
```
tiagolobocastro: Should use correct base for the `i` as mentioned above

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -6,6 +6,9 @@ logLevel: INFO
 # -- Format of the logs (json, pretty)
 logFormat: json
 
+# -- Reserved storage amount on each node (Supports percentage and size) [e.g. `25%` or `50GB` or `10MiB`]
+reservedStorage: 0B
```
tiagolobocastro: So this is reserved but anything other than rawfile?

rawfile/utils/rawfile.py
```diff
@@ -299,6 +299,11 @@ def get_capacity():
     capacity = disk_free_size
     for volume_stat in get_volumes_stats().values():
         capacity -= volume_stat["total"] - volume_stat["used"]
+    reserved_storage = CONFIG.get("reserved_storage", 0)
+    if isinstance(reserved_storage, int):
+        capacity -= CONFIG.get("reserved_storage", 0)
```
tiagolobocastro: IIRC we truncate the image files today, so that means our capacity doesn't actually decrease unless you write to it?
So this reserved wouldn't work if you create a bunch of volumes at once, and write to them later.

deploy/helm/rawfile-localpv/values.yaml
tiagolobocastro: I wonder if it should be the other way around, we specify how much we allow rawfile to take?

mhkarimi1383: As mentioned in the issue sometimes users are having a single drive with single partition, if they want to do that in other ways they should create an exclusive partition for that, but it is not a good option when you have for example 100 GB of storage

rawfile/utils/units.py
mhkarimi1383: We are converting input to upper case

rawfile/utils/rawfile.py
mhkarimi1383: When we truncate we are actually allocating, so we have a file with desired size and the size is reserved for that, if that's not the case, we should change it

rawfile/utils/units.py
mhkarimi1383: Will fix that

rawfile/utils/rawfile.py
tiagolobocastro: truncate just sets the size, see:
```shell
> df -h
/dev/nvme2n1p2  457G  260G  174G  60% /
> truncate -s 4TiB file.io
> df -h
/dev/nvme2n1p2  457G  260G  174G  60% /
```

> we should change it

Depends whether we want to thin provision or not. By default we are thin provisioning with truncate.
We should also add a helm chart and storage class override parameter to make it thick with `fallocate`, ex:
```shell
> df -h
/dev/nvme2n1p2  457G  260G  174G  60% /
> fallocate -l50GiB file.io
> df -h
/dev/nvme2n1p2  457G  310G  124G  72% /
```

mhkarimi1383: Will add this here :+1:

deploy/helm/rawfile-localpv/values.yaml
tiagolobocastro: Yeah it's good to make sure your apps/OS doesn't run out of space.
Just wondering if it's better to use reserved space for OS or for rawfile itself, ie: how much rawfile can take
And with truncated volumes, how do we determine this?

rawfile/utils/units.py
mhkarimi1383: Fixed

deploy/helm/rawfile-localpv/values.yaml
tiagolobocastro: @addyess would like to hear your thoughts here. Is this something you've considered on your use case?

mhkarimi1383:
In thick provisioning we will always report 100% usage on volumes as I understand, it should be changed to get metrics on volumes that are mounted. Not every volume on the node

rawfile/utils/fallocate/build.py
```diff
@@ -0,0 +1,21 @@
+from cffi import FFI
```
tiagolobocastro, line 1: just wondering, is this worth the complication and extra build dependency vs just calling fallocate binary?

rawfile/utils/rawfile.py
```diff
@@ -299,6 +312,11 @@ def get_capacity():
...
```
tiagolobocastro: can this underflow btw?

rawfile/utils/rawfile.py
tiagolobocastro: Ah and above we are already looping the volume list doing total-used, I had actually missed that!
But then this means even with thin provision, rawfile will not use thin provisioning in the capacity?
Maybe we need some more configuration like overcommit? Probably let's not change that here

rawfile/utils/fallocate/build.py
mhkarimi1383: I have done this for safety reasons, while calling binary we can't set fd, So we can't call fsync or whatever

The same reason that we are using truncate function instead of it's binary

rawfile/utils/rawfile.py
mhkarimi1383: You mean if the reserved storage is bigger than capacity?

It can underflow, will make it to return zero if value is lower than zero

rawfile/utils/rawfile.py
mhkarimi1383: I think everything is OK except for volume usages (thick volumes are returning 100% usage every time, So that why I think that we should do usage calculation when volume is I use by it's usage at the loop dev

rawfile/utils/fallocate/build.py
tiagolobocastro: Yeah I get that, but in truncate was an easy change.
Can't we do run("fallocate $file"); run("fsync $file") ?

rawfile/utils/rawfile.py
mhkarimi1383: Fixed

rawfile/utils/fallocate/build.py
mhkarimi1383: We can move this to use commands for `fallocate`, But I would prefer to do that since I don't know how the command is actually working

mhkarimi1383: https://github.com/util-linux/util-linux/blob/master/sys-utils%2Ffallocate.c

Looks like the same as what we are doing (for most of the parts :)), but we could have more control when calling fallocate directly.

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -6,6 +6,11 @@ logLevel: INFO
...
+# -- Overrides total capacity of the storage (Support size values) [e.g. `50GB` or `10MiB`]
+capacityOverride: 0B
```
tiagolobocastro: I don't like having these special cases where 0 means something different than 0 :)

```suggestion
capacityOverride: ""
```

tiagolobocastro: nit: on the naming, how about: `allowedStorage` or `allowedCapacity` ?

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -6,6 +6,11 @@ logLevel: INFO
 # -- Format of the logs (json, pretty)
 logFormat: json
 
+# -- Reserved storage amount on each node (Supports percentage and size) [e.g. `25%` or `50GB` or `10MiB`]
```
tiagolobocastro: I think this to be made clearer here that is storage reserved for the host and not for the rawfile itself

rawfile/utils/rawfile.py
```diff
@@ -49,9 +51,14 @@ class AccessType(Enum):
 
 def path_stats(path):
     fs_stat = os.statvfs(path)
+
+    total = CONFIG.get("capacity_override", 0) or (fs_stat.f_frsize * fs_stat.f_blocks)
```
tiagolobocastro: Not sure if we can do this here, path_stats may be called for a volume rather than the data_dir for volumes.
In such case we should report fs_avail of the volume itself?

tiagolobocastro: Also should we truncate this to the actual available capacity on the node?

deploy/helm/rawfile-localpv/values.yaml
tiagolobocastro: @Abhinandan-Purkait @niladrih @avishnu any thoughts here?
My main concern is that this `reservedStorage` is a moving target and it doesn't work with thin provisioned volumes;
I'd rather have the opposite, ex something like this:
- usableStorage: How much rawfile is allowed to use
- overCommitment: How much rawfile is allowed to overcommit beyond the usableStorage for thin volumes

tiagolobocastro: Or maybe it doesn't matter, and we'd always get into trouble with the thin provisioned volumes anyway, if the host writes to the fs after we've created the thin volumes since we can't actually reserve storage just for rawfile anyway...

rawfile/utils/rawfile.py
mhkarimi1383: I will fix that now

mhkarimi1383: Fixed. It will apply only for data dir

deploy/helm/rawfile-localpv/values.yaml
mhkarimi1383: It is up to users itself, but having this option will help users to be able to their OSes in a safe place then they only have one Storage Partition on their nodes

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -6,6 +6,11 @@ logLevel: INFO
 # -- Format of the logs (json, pretty)
 logFormat: json
 
+# -- Reserved storage amount on each node for data dir storage on each host (Supports percentage and size) [e.g. `25%` or `50GB` or `10MiB`]
```
tiagolobocastro: Also, this is currently the reserved from the free space right, not the total host fs size?

tiagolobocastro: hmm would something like `reservedCapacity` be a better name to keep in-line with the `capacityOverride` ?

mhkarimi1383: It is for total size right now, if you check out the impl.

mhkarimi1383: Will change the name to be more consistent

mhkarimi1383: Done.

tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

PR #162, opened 2025-07-11
:technologist: Add helm-docs pre-commit hook

mhkarimi1383:


mhkarimi1383:
`--skip-version-footer` has been added because the pre-commit was ignoring me when I set that to false and I don't know why it is enable in their pre-commit hook and how it is set to default, I found nothing on their repo

tiagolobocastro (approved):
Do we run helm-docs on CI? If not we should in case someone skips pre-commit checks

mhkarimi1383:
Yeah, we can remove helm docs from actions

mhkarimi1383:
I have removed helm docs from ci and actions

tiagolobocastro:
> Yeah, we can remove helm docs from actions

No sorry I was asking if it was already there
Should not remove it, because someone will skip the pre-commit check, guaranteed :)

mhkarimi1383:
> > Yeah, we can remove helm docs from actions
> 
> No sorry I was asking if it was already there Should not remove it, because someone will skip the pre-commit check, guaranteed :)

We have that in pre-commit ci and in runs on every PR, I will bring it back if I'm not getting it correctly

tiagolobocastro:
> > > Yeah, we can remove helm docs from actions
> > 
> > 
> > No sorry I was asking if it was already there Should not remove it, because someone will skip the pre-commit check, guaranteed :)
> 
> We have that in pre-commit ci and in runs on every PR, I will bring it back if I'm not getting it correctly

Oh right!
Yeah give it a test by pushing a bad readme and see if it's flagged
Though we need to add the pre-commit as a required CI pass

mhkarimi1383:
I have tested this :)
That's why there are force pushes there 😁

mhkarimi1383:
@tiagolobocastro Can you approve again? :)

tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

PR #28, opened 2024-02-25
♻️  refactor: some chart improvments and optimizations

mhkarimi1383:
Dear @semekh 

Excuse me for making lots of changes in a single PR, But I tried to document change here and in my commit description

making images customizable
adding StorageClass management via helm chart
move to `registry.k8s.io` for kubernetes images (all of the images are available there)
switch from yaml anchors/pointers to global values and helpers to handle default values so it will make customizations easier for resources/images
getting default tag from appVersion in Chart.yaml
dataDir path for plugin is customizable now (closes #3)
documenting chart using helm-docs tool (with comments in values file)

Also I wanted to add VolumeSnapshotClasses but I don't know if there is any parameter required to make snapshotclass working,
if you give me examples I will add it too

As for chart changes I don't know it I have to bump major or minor

If it's OK I want to also add GitHub actions to use GH-Pages for Helm chart publish and removing the need of cloning git repo to install the chart

PR #164, opened 2025-07-13
:zap: Tuning gRPC server

mhkarimi1383:
Improve gRPC server to only server Services when needed and set number of workers to 1 to match other provisioners

### DIFF
--- a/rawfile/rawfile.py
+++ b/rawfile/rawfile.py
@@ -61,18 +61,34 @@ def cli(
 @click.option(
     "--enable-metrics/--disable-metrics", envvar="ENABLE_METRICS", default=True
 )
-def csi_driver(endpoint, nodeid, enable_metrics, metrics_port):
-    migrate_all_volume_schemas()
+@click.option("--plugin-type", envvar="PLUGIN_TYPE")
+@click.option("--grpc-workers", envvar="GRPC_WORKERS", default="10")
+def csi_driver(
+    endpoint, nodeid, enable_metrics, metrics_port, plugin_type, grpc_workers
+):
     if enable_metrics:
         expose_metrics(nodeid, metrics_port)
-    server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
+    logger.debug(
+        "Starting gRPC server",
+        endpoint=endpoint,
+        nodeid=nodeid,
+        enable_metrics=enable_metrics,
+        plugin_type=plugin_type,
+    )
+    server = grpc.server(futures.ThreadPoolExecutor(max_workers=int(grpc_workers)))
     csi_pb2_grpc.add_IdentityServicer_to_server(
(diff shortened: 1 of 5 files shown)


mhkarimi1383:
@tiagolobocastro snappshotter should be in controller and we should use remote fn to run them on nodes

it may pick a wrong Resource from a wrong node

tiagolobocastro:
> @tiagolobocastro snappshotter should be in controller and we should use remote fn to run them on nodes
> 
> it may pick a wrong Resource from a wrong node

we're running in a different mode than usual here, where controller actually runs on the node
I think I have a diagram somewhere here in a github comment

also for snapshots we're running the distributed snapshotting for local volumes: <https://github.com/kubernetes-csi/external-snapshotter?tab=readme-ov-file#distributed-snapshotting>

mhkarimi1383:
But snapshot methods are in the Controller part, I have lvm local pv, and it was in the controller

tiagolobocastro:
> But snapshot methods are in the Controller part, I have lvm local pv, and it was in the controller

Yeah we run the unified deployment mode, it's not the same as lvm local pv.
I've mentioned this before: <https://github.com/openebs/rawfile-localpv/pull/28#discussion_r1974043841>

mhkarimi1383:
Will check it out

mhkarimi1383:
Since this is a thing in csi proto and I can't change it I will leave node plugin to serve controller things too for now

rawfile/rawfile.py
```diff
@@ -61,18 +61,31 @@ def cli(
...
         expose_metrics(nodeid, metrics_port)
-    server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
+    logger.debug(
+        "Starting gRPC server",
+        endpoint=endpoint,
+        nodeid=nodeid,
+        enable_metrics=enable_metrics,
+        plugin_type=plugin_type,
+    )
+    server = grpc.server(futures.ThreadPoolExecutor(max_workers=1))
```
tiagolobocastro: why is this being modified to 1 worker?

mhkarimi1383: It was to match others, but We should have a better value, I have the same question that why it is 10 right now

tiagolobocastro: others as in?
I'm not sure exactly why 10 here, but on mayastor for example we have a default value of 10 which allows up to 10 concurrent csi operations

mhkarimi1383: I saw LVM I think

mhkarimi1383: I will default it to 10 then I will make it configurable

tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio squash

mhkarimi1383:
@Mergifyio squash

mhkarimi1383:
@Mergifyio queue

PR #168, opened 2025-07-15
:sparkles: Add format options support

mhkarimi1383:


rawfile/rawfile_servicer.py
```diff
@@ -204,14 +204,14 @@ def CreateVolume(self, request, context):
...
-                k.lower(): v for k, v in request.parameters.items()
-            }
+            params = normalize_parameters(request.parameters.items())
             thin_provision = params.get("thinprovision", "no")
+            format_options = params.get("formatoptions", "")
             init_rawfile(
                 volume_id=request.name,
                 size=size,
                 thin_provision=str_to_bool(thin_provision),
+                format_options=format_options.split(" "),
```
tiagolobocastro: we don't need to add to the metadata, we can simply return them on the `volume_context` and then we receive the context in the stage call

mhkarimi1383: Yeah, just like lvm local pv

tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

PR #173, opened 2025-07-19
:zap: Improve volume verifier and task pod to be more safe

mhkarimi1383:
Just to make it easier to work with, and having more strict types than reading raw string and formatting from file

Also added a simple debug log to log the manifest of the created pod


tiagolobocastro (approved):

mhkarimi1383:
@Mergifyio queue

Issue #111, opened 2025-05-28
Add analytics in line with other local engines


tiagolobocastro:
### Describe the problem/challenge you have

No analytics

### Describe the solution you'd like

Add analytics in line with the other local engines

tiagolobocastro:
Completed by #172 and #174

PR #174, opened 2025-07-19
📊📈 - OpenEBS analytics using Ga4

tiagolobocastro:
feat(analytics): add openebs analytics

    Adds a configmap allowing us to track installs
    Adds 24h ping
    Adds volume provision events

---

    fix(resize): set timeout to 30s
    
    Allows for a longer time spend in the resizer pod, which can happen
    as the pod is quite cpu limited by default.

---

    fix(k8s): try to load in-cluster first
    
    Removes the weird error message about the config file

---

    fix: set provisioner version from metadata

---

    fix(analytics): set the actual hostname

    This ensures the certificate validation is done with correct hostname
    and not the resolved ip address.

---

    docs: fix a few typos


rawfile/analytics/ga4/usage.py
```diff
@@ -1,18 +1,38 @@
...
+        self.send()
+
+    def send_ping(self):
+        self.builder.category("ping")
+        self.send()
+
+    def volume_provision(self, name, capacity):
+        self.builder.category("volume_provision").volume_name(name).volume_capacity(
+            capacity
+        ).replica_count(1)
```
niladrih: On the localpvs, we're doing [this](https://github.com/openebs/google-analytics-4/blob/develop/usage/const.go#L21). We're doing that because the older library did that as well. It's a string which says 'replica:1'. There's no need to make any change to what you're doing here, just a heads up.

niladrih (approved):

rawfile/rawfile.py
```diff
@@ -26,6 +27,10 @@
 @click.option("--log-level", envvar="LOG_LEVEL", default="INFO")
 @click.option("--reserved-capacity", envvar="RESERVED_CAPACITY", default="0")
 @click.option("--capacity-override", envvar="CAPACITY_OVERRIDE", default="0")
+@click.option("--ga-enabled/--ga-disabled", is_flag=True)
+@click.option("--ga-id", envvar="GA_ID", default="G-LPD03CGEVN")
+@click.option("--ga-key", envvar="GA_KEY", default="ibj6b0nxRqC5ti4xzUgRow")
```
niladrih: These are the values to the actual openebs property: https://github.com/openebs/google-analytics-4/blob/develop/usage/const.go#L5

Does this need to be disabled CI?

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -20,6 +20,8 @@ global:
   imagePullPolicy: IfNotPresent
   # -- Default image pull secret for images
   imagePullSecrets: []
+  analytics:
+    enabled: true
```
mhkarimi1383, line 25: Please add docs for this option

rawfile/analytics/ga4/__init__.py
```diff
@@ -1,6 +1,81 @@
...
+        task_queue.put(event_fn)
+
+
+def run_event_worker():
+    if enabled():
+        thread = threading.Thread(target=worker, daemon=True)
+        thread.start()
+
+
+def enabled() -> bool:
```
mhkarimi1383, line 78:
```suggestion
 enabled = lambda: bool(CONFIG["ga_enabled"]) # cast to ensures type hints
```

rawfile/rawfile_servicer.py
```diff
@@ -221,6 +222,11 @@ def CreateVolume(self, request, context):
             else:
                 raise exc
 
+        def volume_provision(usage: Usage):
+            usage.volume_provision(request.name, size)
+
+        send_event(volume_provision)
```
mhkarimi1383, line 230: in `send_event` calls should we ignore errors, So CSI call wont don't get error?

tiagolobocastro: There should not be any errors on `send_event` since we only push to the queue?

rawfile/analytics/ga4/__init__.py
tiagolobocastro: Done

deploy/helm/rawfile-localpv/values.yaml
tiagolobocastro: Done

rawfile/rawfile.py
tiagolobocastro: Yes I think we should disable on the CI
It's only enabled now so I can check it works, once I'm happy it's working I'll both disable it on CI and set the actual ga properties

rawfile/analytics/ga4/usage.py
tiagolobocastro: Example so it'd be this?
```json
....,
"vol_replica_count": "replica:1"
```

rawfile/analytics/ga4/__init__.py
```diff
@@ -1,6 +1,81 @@
...
+            pass
+        task_queue.task_done()
+    logger.info("Event Worker Exited")
+
+
+def shutdown_event_worker():
+    stop_event.set()
+
+
+def send_event(event_fn):
```
mhkarimi1383, line 67: send_event = lambda event_fn : task_queue.put(event_fn)

rawfile/analytics/ga4/ping.py
```diff
@@ -0,0 +1,111 @@
...
+        if not self.installed:
+            self.usage.send_install()
+            self.installed = True
+
+        write_config_map(self.configmap, self.install_ts_key, self.install_ts)
+        # After we've sent the install event, and written it to the cm, we can now start the ping
+        return self.install_ts
+
+
+def time_to_str(tm) -> str:
```
mhkarimi1383, line 84:
```suggestion
time_to_str = lambda tm : str(tm.strftime(time_fmt))
```

rawfile/analytics/ga4/ping.py
```diff
@@ -0,0 +1,111 @@
...
+    return tm.strftime(time_fmt)
+
+
+def now_to_str() -> str:
```
mhkarimi1383, line 88:
```suggestion
now_to_str = lambda : datetime.now(timezone.utc).strftime(time_fmt)
def now_to_str() -> str:
```

rawfile/rawfile_servicer.py
mhkarimi1383: Oh, I have not seen that

rawfile/analytics/ga4/__init__.py
```diff
@@ -1,6 +1,81 @@
...
+            task(event_usage)
+        except Exception as e:
+            # todo retry?
+            logger.error("Event Worker Error", exception=e)
+            pass
+        task_queue.task_done()
+    logger.info("Event Worker Exited")
+
+
+def shutdown_event_worker():
```
mhkarimi1383, line 63:
```suggestion
shutdown_event_worker = lambda : stop_event.set()
def shutdown_event_worker():
```

rawfile/analytics/ga4/__init__.py
tiagolobocastro: the linter complains about these lambda, asking me to add them as def (same thing on the enable)

mhkarimi1383: Ok, no problem

I have suggested them since they where some simple functions or aliases

mhkarimi1383 (approved):

rawfile/analytics/ga4/__init__.py
tiagolobocastro: Just for completeness, adding here:
```
rawfile/analytics/ga4/__init__.py:52:1: E731 Do not assign a `lambda` expression, use a `def`
   |
52 | shutdown_event_worker = lambda : stop_event.set()
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ E731
   |
   = help: Rewrite `shutdown_event_worker` as a `def`
```

rawfile/analytics/ga4/event.py
```diff
@@ -1,66 +1,103 @@
...
+
+
 class OpenEBSEventBuilder:
     def __init__(self):
         self.params = {}
+        self.project("OpenEBS").engine_name("rawfile-localpv").engine_version(
+            PROVISIONER_VERSION
+        )
 
+    # Specify Project name, ex OpenEBS
```
mhkarimi1383: Can you please make them to be docstring instead of comments?

mhkarimi1383 (approved):

mhkarimi1383 (approved):

rawfile/analytics/ga4/usage.py
Abhinandan-Purkait: No, `vol_replica_count` is supposed to be 1, there are 3 fields which are common in every custom event, `event_category`, `event_value`, `event_action`, these 3 need to be present in every custom event + whatever field we want.

So for volume creation, the data would be like this,

1. `event_action`: `replica:1`
2. `event_category`: `volume_provision`
3. `event_value`: `50` (size in iBytes without the unit, just the integer, this is because then in UI it automatically adds it up to show total provision in a day)
4. `vol_replica_count`: `1` --> Custom field for replica count
5. `vol_capacity`: `50Gi` --> Custom field for capacity

The first 3 fields are special as they help up formulate charts later.

Abhinandan-Purkait: Its important that the `event_value` is always an integer,

for install and ping events its number of nodes, for provision and deprovision its the size in iBytes with only the integer part.

rawfile/analytics/ga4/usage.py
```diff
@@ -1,18 +1,54 @@
...
+
+    def volume_deprovision(self, name: str):
+        self.builder.category("volume-deprovision").volume_name(name)
+        self.send()
 
     def send(self):
         event = self.builder.build()
-        self.client.send_event("openebs_event", event)
+        category = event["event_category"] if "event_category" in event else "unknown"
+        engine_name = event["engine_name"] if "engine_name" in event else "unknown"
```
Abhinandan-Purkait: Not sure if we should send if it's unknown without the category we can't really formulate any chart/graph, although the event count would be up.

rawfile/analytics/ga4/usage.py
```diff
@@ -1,18 +1,54 @@
...
+        self.send()
+
+    def send_ping(self):
+        self.builder.category("rawfile-ping")
+        self.send()
+
+    def volume_provision(self, name: str, capacity: int):
+        self.builder.category("volume-provision").volume_name(name).volume_capacity(
+            capacity
+        ).replica_count(1)
```
Abhinandan-Purkait: How are the other fields set here? I couldn't find it here?

Abhinandan-Purkait: Like, `event_action` and `event_value`, etc

rawfile/consts.py
```diff
@@ -11,3 +14,5 @@
 VOLUME_IN_USE_EXIT_CODE = 102
 VOLUME_IS_ATTACHED = "openebs.io/rawfile/is-attached"
 FORMAT_OPTIONS_KEY = "openebs.io/rawfile/format-options"
+GA_ID = "Ry1MUEQwM0NHRVZO"
+GA_KEY = "aWJqNmIwbnhScUM1dGk0eHpVZ1Jvdw=="
```
Abhinandan-Purkait: Is this encoded?

tiagolobocastro: Yes it is

rawfile/analytics/ga4/usage.py
tiagolobocastro: Discussed offline, the `event_*` are not used in the openebs analytics atm

rawfile/rawfile.py
tiagolobocastro: Disabled on CI

rawfile/analytics/ga4/usage.py
tiagolobocastro: Raising an exception now if not present

tiagolobocastro:
@mergifyio queue

tiagolobocastro:
Resolves #111

PR #182, opened 2025-07-22
:bug: Fix GA4 ping timedelta parse

mhkarimi1383:
it should not parse anymore since we are using pydantic and it is parsing the value automatically


tiagolobocastro (approved):

mhkarimi1383:
@mergifyio queue

PR #189, opened 2025-07-30
Nix update and changelog

tiagolobocastro:
Updates nixpkgs to track the stable branch
Update changelog with the previous release date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ---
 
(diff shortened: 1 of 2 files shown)


mhkarimi1383:
I think we should also change our nix actions to use official ones or one from nix community
(We could also have caching there as well)

mhkarimi1383 (approved):

tiagolobocastro:
> I think we should also change our nix actions to use official ones or one from nix community (We could also have caching there as well)

yes we should do that

@mergifyio queue

tiagolobocastro:
@mergifyio queue

PR #190, opened 2025-07-31
:green_heart: Change nix action to use cachix and add nix-cache

mhkarimi1383:


.github/workflows/helm-chart.yml
```diff
@@ -17,12 +17,18 @@ jobs:
...
-          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
-          nix-shell deploy/helm/shell.nix --run "echo"
+          echo "nix_path=nixpkgs=$(jq -r '.nixpkgs.url' nix/sources.json)" >> $GITHUB_OUTPUT
+      - uses: cachix/install-nix-action@v25
+        with:
+          nix_path: "${{ steps.nixpath.outputs.nix_path }}"
+      - name: Setup nix cache
+        uses: cachix/cachix-action@v14
+        with:
+          name: nix-cache
```
tiagolobocastro: Doesn't this have to be the name of our cache in cachix?
Also we'd to specify the token in order to write?
Do they even have a free tier for OSS?

mhkarimi1383: Yeah, I was thinking that it stores them in GitHub Actions cache, but it is using the cachix itself, it is free for OSS softwares

tiagolobocastro: Ah the magic nix cache has now been restored, so we could use it again

tiagolobocastro: Actually, do we even need the cache here? We're not building our own derivations

mhkarimi1383: No we don't need, I have added this and I was thinking that it stores cache in GitHub Actions

tiagolobocastro: Let's remove it then, we can circle back if we ever build our own derivations.

mhkarimi1383: Will remove that right now

tiagolobocastro (approved):

mhkarimi1383:
@mergifyio queue

PR #211, opened 2025-08-27
:green_heart: Add nix cache on CI using github actions

mhkarimi1383:


.github/workflows/helm-chart.yml
```diff
@@ -16,14 +16,23 @@ jobs:
...
         with:
           nix_path: "${{ steps.nixpath.outputs.nix_path }}"
+          install_options: --no-daemon
+          extra_nix_config: |
+            keep-env-derivations = true
+            keep-outputs = true
+      - name: Restore and save Nix store
+        uses: nix-community/cache-nix-action@v6
+        with:
+          primary-key: helm-nix-${{ hashFiles('deploy/helm/shell.nix', 'nix/*') }}
```
tiagolobocastro, line 34: how is garbage collection done?

mhkarimi1383: Post Restore phase:

    The cache-nix-action purges caches whose keys are the same as the primary key and that were created or last accessed more than a given time ago (inputs: purge, purge-created, purge-last-accessed, purge-primary-key).

    When there's no cache whose key is the same as the primary key, the cache-nix-action collects garbage in the Nix store and saves a new cache (inputs: save, gc-max-store-size, gc-max-store-size-linux, gc-max-store-size-macos).

    The cache-nix-action purges caches whose keys match some of the given prefixes in a given list of key prefixes and that were created or last accessed more than a given time ago relative to the start of the Post Restore phase (purge, purge-prefixes, purge-created, purge-last-accessed, purge-primary-key).

mhkarimi1383: it has been done at Post Restore step before updating the cache

tiagolobocastro (approved):

mhkarimi1383:
@mergifyio queue

PR #213, opened 2025-08-30
:sparkles: Complete volume snapshot support and better snapshot management 

### LINKED ISSUE #19: Add XFS snapshot support via reflink
Looks lke XFS added reflink support, this can be leveraged to support snapshots for XFS

https://blogs.oracle.com/linux/post/xfs-data-block-sharing-reflink

mhkarimi1383:
Resolves #19

### DIFF
--- a/docs/design/taskmanager.md
+++ b/docs/design/taskmanager.md
@@ -1,0 +1,96 @@
+# Task Manager Design Document
+
+## System Overview
+
+The Task Manager is a robust system for handling asynchronous task execution with state persistence, retry capabilities, and graceful shutdown handling. It's currently integrated with snapshot management functionality.
+
+## Core Components
+
+### 1. Task States
+
+```python
+class TaskState(StrEnum):
+    PENDING = "Pending"
+    RUNNING = "Running"
+    FAILED = "Failed"
+    COMPLETED = "Completed"
+```
+
+### 2. Task Information Structure
+
+```python
+class TaskInfo(TypedDict):
+    task: str          # Task name/identifier
+    args: list         # Positional arguments
+    kwargs: dict       # Keyword arguments
+    retry_count: int   # Number of retry attempts
+    state: TaskState   # Current state of task
+```
+
+### 3. Key Features
+
+#### State Management
+
+- **Persistent Storage**
+  - Tasks stored in JSON format at `{DATA_DIR}/tasks.json`
+  - Thread-safe operations using Lock mechanism
+  - Automatic state recovery after crashes
+  - Handles stuck tasks by marking them as failed
+
+#### Task Execution
+
+- Asynchronous execution using provided executor
+- Task identification using MD5 hash of task parameters
+- Current supported tasks:
+
+  ```python
+  class TaskName(StrEnum):
+      CREATE_SNAPSHOT = "CreateSnapshot"
+  ```
+
+#### Retry Mechanism
+
+- Configurable retry settings:
+  - Retry interval (default: 5 seconds)
+  - Maximum retries (default: 5 attempts)
+- Automatic retry for failed tasks
+- Cleanup of non-retriable and completed tasks
+
+## Technical Implementation
+
+### 1. Thread Safety
+
+- Uses `threading.Lock` for file operations
+- Ensures atomic operations on task storage
+
+### 2. Task Lifecycle
+
+1. Task submission
+2. State transitions: PENDING → RUNNING → COMPLETED/FAILED
+3. Automatic retry for failed tasks
+4. Cleanup of completed/non-retriable tasks
+
+### 3. Shutdown Process
+
+- Graceful shutdown with timeout
+- Marks running tasks as failed
+- Cancels pending futures
+- Preserves task states
+
+## Error Handling
+
+- Custom `TaskManagerShuttingDown` exception
+- Automatic crash recovery
+- Task state preservation
+- Exception logging with context
+
+## Usage Example
+
+```python
+task_manager = TaskManager(executor, retry_interval=5, max_retry=5)
+task_id = task_manager.run_task(
+    TaskName.CREATE_SNAPSHOT,
+    *args,
+    **kwargs
+)
+```
--- a/rawfile/bd2fs.py
+++ b/rawfile/bd2fs.py
@@ -7,36 +7,47 @@ import time
 from csi import csi_pb2, csi_pb2_grpc
 from csi.csi_pb2 import (
     CreateVolumeRequest,
+    ListSnapshotsResponse,
     NodeExpandVolumeRequest,
     NodePublishVolumeRequest,
     NodeStageVolumeRequest,
     NodeUnpublishVolumeRequest,
     NodeUnstageVolumeRequest,
+    Snapshot,
 )
 from declarative import (
     mount,
     unmount,
 )
-from utils.rawfile import be_absent
+from utils import task_manager
+from utils.rawfile import (
+    attach_loop,
+    be_absent,
+    detach_loops,
+    gc_if_needed,
+    img_file,
+    metadata,
+)
 from google.protobuf.timestamp_pb2 import Timestamp
 from utils.logs import log_grpc_request
 from filesystem import get_from_device_or_fallback, from_device
 from filesystem.utils import get_device_for_mountpoint
 from rawfile_servicer import check_access_type, get_access_type
 from utils.rawfile import (
-    img_file,
-    attach_loop,
     AccessType,
-    detach_loops,
 )
 from utils.devices import path_stats
 from filesystem.base import UnknownFileSystemError
 from utils.lock import VolLock
+from utils.task_manager import TaskManager
+from utils.snapshot_manager import manager as snapshot_manager
+import consts
 
 
 class Bd2FsIdentityServicer(csi_pb2_grpc.IdentityServicer):
-    def __init__(self, bds: csi_pb2_grpc.IdentityServicer):
+    def __init__(self, bds: csi_pb2_grpc.IdentityServicer, task_manager: TaskManager):
         self.bds = bds
+        self._task_manager = task_manager
 
     @log_grpc_request
     def GetPluginInfo(self, request, context):
@@ -52,8 +63,9 @@ class Bd2FsIdentityServicer(csi_pb2_grpc.IdentityServicer):
 
 
 class Bd2FsNodeServicer(csi_pb2_grpc.NodeServicer):
-    def __init__(self, bds: csi_pb2_grpc.NodeServicer):
+    def __init__(self, bds: csi_pb2_grpc.NodeServicer, task_manager: TaskManager):
         self.bds = bds
+        self._task_manager = task_manager
 
     # @log_grpc_request
     def NodeGetCapabilities(self, request, context):
@@ -223,15 +235,16 @@ class Bd2FsNodeServicer(csi_pb2_grpc.NodeServicer):
 
 
 class Bd2FsControllerServicer(csi_pb2_grpc.ControllerServicer):
-    def __init__(self, bds: csi_pb2_grpc.ControllerServicer):
+    def __init__(self, bds: csi_pb2_grpc.ControllerServicer, task_manager: TaskManager):
         self.bds = bds
+        self._task_manager = task_manager
 
     @log_grpc_request
     def ControllerGetCapabilities(self, request, context):
         return self.bds.ControllerGetCapabilities(request, context)
 
     @log_grpc_request
-    def CreateVolume(self, request, context):
+    def CreateVolume(self, request: CreateVolumeRequest, context):
         if len(request.volume_capabilities) != 1:
             context.abort(
                 grpc.StatusCode.INVALID_ARGUMENT, "Exactly one cap is supported"
@@ -252,16 +265,14 @@ class Bd2FsControllerServicer(csi_pb2_grpc.ControllerServicer):
         access_type = volume_capability.WhichOneof("access_type")
         check_access_type(access_type)
 
-        bd_request = CreateVolumeRequest()
-        bd_request.CopyFrom(request)
-        bd_request.capacity_range.required_bytes = max(
+        request.capacity_range.required_bytes = max(
             request.capacity_range.required_bytes, 10 * 1024 * 1024
         )  # At least 10MB
 
         # FIXME: update access_type
         # bd_request.volume_capabilities[0].block = ""
         # bd_request.volume_capabilities[0].mount = None
-        return self.bds.CreateVolume(bd_request, context)
+        return self.bds.CreateVolume(request, context)
 
     @log_grpc_request
     def DeleteVolume(self, request, context):
@@ -284,55 +295,125 @@ class Bd2FsControllerServicer(csi_pb2_grpc.ControllerServicer):
     @log_grpc_request
     def CreateSnapshot(self, request: csi_pb2.CreateSnapshotRequest, context):
         with VolLock(request.source_volume_id):
-            file = img_file(request.source_volume_id)
-            loop_dev = attach_loop(file)
-            fs = from_device(loop_dev)
-            if not fs:
-                raise UnknownFileSystemError(
-                    device=loop_dev, volume_id=request.source_volume_id
-                )
-            fs.create_snapshot(name=request.name)
-
-            if fs.mountpoint is None:
-                detach_loops(file)
-
-            creation_time_ns = time.time_ns()
-            snapshot_id = f"{request.source_volume_id}/{request.name}"
-            # TODO: detach loopdev when we get seperated loop devices
-            nano = 10**9
-            return csi_pb2.CreateSnapshotResponse(
-                snapshot=csi_pb2.Snapshot(
-                    size_bytes=0,
-                    snapshot_id=snapshot_id,
-                    source_volume_id=request.source_volume_id,
-                    creation_time=Timestamp(
-                        seconds=creation_time_ns // nano, nanos=creation_time_ns % nano
-                    ),
-                    ready_to_use=True,
-                )
+            volume_meta = metadata(request.source_volume_id)
+            freezefs = volume_meta.get("freezefs", False)
+            copy_on_write_param = volume_meta.get("copy_on_write", None)
+            copy_on_write = (
+                copy_on_write_param
+                if copy_on_write_param is not None
+                else consts.COW_SUPPORTED
             )
+
+        def _get_current_snapshot():
+            _current = snapshot_manager.list_snapshots(
+                volume_id=request.source_volume_id, snapshot_name=request.name
+            )
+            if not _current.data:
+                return None
+            return _current.data[0]
+
+        nano = 10**9
+        current = _get_current_snapshot()
+        if not current:
+            self._task_manager.run_task(
+                task_manager.TaskName.CREATE_SNAPSHOT,
+                volume_id=request.source_volume_id,
+                name=request.name,
+                copy_on_write=copy_on_write,
+                freeze_fs=freezefs,
+            )
+        start_time = time.time()
+        while time.time() - start_time < 30:
+            current = _get_current_snapshot()
+            if current and current.ready:
+                return csi_pb2.CreateSnapshotResponse(
+                    snapshot=csi_pb2.Snapshot(
+                        size_bytes=current.size_bytes,
+                        snapshot_id=current.snapshot_id,
+                        source_volume_id=current.volume_id,
+                        ready_to_use=current.ready,
+                        creation_time=Timestamp(
+                            seconds=int(current.creation_time),
+                            nanos=int((current.creation_time % 1) * nano),
+                        ),
+                    )
+                )
+            time.sleep(0.5)
+        creation_time_ns = time.time_ns()
+        snapshot_id = f"{request.source_volume_id}/{request.name}"
+        return csi_pb2.CreateSnapshotResponse(
+            snapshot=csi_pb2.Snapshot(
+                size_bytes=0,
+                snapshot_id=snapshot_id,
+                source_volume_id=request.source_volume_id,
+                ready_to_use=False,
+                creation_time=Timestamp(
+                    seconds=creation_time_ns // nano,
+                    nanos=creation_time_ns % nano,
+                ),
+            )
+        )
 
     @log_grpc_request
     def DeleteSnapshot(self, request: csi_pb2.DeleteSnapshotRequest, context):
         snapshot_id = request.snapshot_id
         volume_id, name = snapshot_id.rsplit("/", 1)
+        if (
+            len(
+                snapshot_manager.list_snapshots(
+                    volume_id=volume_id, snapshot_name=name
+                ).data
+            )
+            < 0
+        ):
+            with VolLock(volume_id):
+                file = img_file(volume_id)
 
+                loop_dev = attach_loop(file)
+                fs = from_device(loop_dev)
+                if not fs:
+                    raise UnknownFileSystemError(device=loop_dev, volume_id=volume_id)
+                fs.delete_snapshot(name=name)
+
+                if fs.mountpoint is None:
+                    detach_loops(file)
+
+                return csi_pb2.DeleteSnapshotResponse()
+        snapshot_manager.delete_snapshot(volume_id, name)
+        gc_if_needed(volume_id)
+        return csi_pb2.DeleteSnapshotResponse()
+
+    @log_grpc_request
+    def ListSnapshots(self, request: csi_pb2.ListSnapshotsRequest, context):
+        volume_id, name = None, None
+        if request.snapshot_id:
+            snapshot_id = request.snapshot_id
+            volume_id, name = snapshot_id.rsplit("/", 1)
+        offset = None
         try:
-            if not img_file(volume_id).exists():
-                raise FileNotFoundError
-        except FileNotFoundError:
-            return csi_pb2.DeleteSnapshotResponse()
-
-        with VolLock(volume_id):
-            file = img_file(volume_id)
-
-            loop_dev = attach_loop(file)
-            fs = from_device(loop_dev)
-            if not fs:
-                raise UnknownFileSystemError(device=loop_dev, volume_id=volume_id)
-            fs.delete_snapshot(name=name)
-
-            if fs.mountpoint is None:
-                detach_loops(file)
-
-            return csi_pb2.DeleteSnapshotResponse()
+            offset = int(request.starting_token) if request.starting_token else None
+        except ValueError:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, "invalid starting_token")
+        limit = request.max_entries or None
+        snapshot_list = snapshot_manager.list_snapshots(
+            volume_id=volume_id, snapshot_name=name, offset=offset, limit=limit
+        )
+        nano = 10**9
+        return csi_pb2.ListSnapshotsResponse(
+            entries=[
+                ListSnapshotsResponse.Entry(
+                    snapshot=Snapshot(
+                        size_bytes=snapshot.size_bytes,
+                        snapshot_id=snapshot.snapshot_id,
+                        source_volume_id=snapshot.volume_id,
+                        creation_time=Timestamp(
+                            seconds=int(snapshot.creation_time),
+                            nanos=int((snapshot.creation_time % 1) * nano),
+                        ),
+                        ready_to_use=snapshot.ready,
+                    )
+                )
+                for snapshot in snapshot_list.data
+            ],
+            next_token=str(snapshot_list.next_token),
+        )
--- a/rawfile/filesystem/base.py
+++ b/rawfile/filesystem/base.py
@@ -1,10 +1,16 @@
 from abc import ABC, ABCMeta, abstractmethod
+from warnings import deprecated
+from utils.logs import logger
 import subprocess
 from .utils import get_device_for_mountpoint, get_device_fs
 from pathlib import Path
 from typing import Self
 from utils.commands import run
 from utils.errors import InvalidDeviceForMountpointError
+
+fs_snapshot_deprecated = deprecated(
+    "Filesystem level snapshots are not supported anymore"
+)
 
 
 class NotSupportedError(NotImplementedError):
@@ -54,6 +60,22 @@ class FileSystemMountError(FileSystemOperationError):
     operation = "mount"
 
 
+class FileSystemFreezeError(FileSystemOperationError):
+    """
+    Exception raised when filesystem freeze has been failed.
+    """
+
+    operation = "freeze"
+
+
+class FileSystemUnFreezeError(FileSystemOperationError):
+    """
+    Exception raised when filesystem freeze has been failed.
+    """
+
+    operation = "unfreeze"
+
+
 class FileSystemUnmountError(FileSystemOperationError):
     """
     Exception raised when filesystem unmount has been failed.
@@ -70,17 +92,10 @@ class FileSystemResizeError(FileSystemOperationError):
     operation = "resize"
 
 
-class FileSystemCreateSnapshotError(FileSystemOperationError):
-    """
-    Exception raised when filesystem resize has been failed.
-    """
-
-    operation = "create snapshot"
-
-
+@fs_snapshot_deprecated
 class FileSystemDeleteSnapshotError(FileSystemOperationError):
     """
-    Exception raised when filesystem resize has been failed.
+    Exception raised when filesystem delete snapshot has been failed.
     """
 
     operation = "delete snapshot"
@@ -149,25 +164,6 @@ class FileSystem(ABC, metaclass=ABCMeta):
             if not (hasattr(subclass, method) and callable(getattr(subclass, method))):
                 return False
         return True
-
-    def create_snapshot(self, name: str):
-        """
-        Create a snapshot of the current state of the file system.
-        """
-        raise NotSupportedError("create_snapshot")
-
-    def delete_snapshot(self, name: str):
-        """
-        Delete the selected snapshot of the file system.
-        """
-        raise NotSupportedError("delete_snapshot")
-
-    def restore_snapshot(self, name: str):
-        # FIXME: We will probably have some other parameters
-        """
-        Delete the selected snapshot of the file system.
-        """
-        raise NotSupportedError("restore_snapshot")
 
     def format_fs(self, options: list[str] = []) -> str | None:
         """
@@ -252,6 +248,13 @@ class FileSystem(ABC, metaclass=ABCMeta):
             raise FileSystemUnmountError.from_exc(e, self.__filesystem__)
         return output
 
+    @fs_snapshot_deprecated
+    def delete_snapshot(self, name: str):
+        """
+        Delete the selected snapshot of the file system.
+        """
+        raise NotSupportedError("delete_snapshot")
+
     @abstractmethod
     def resize(self) -> str | None:
         """
@@ -272,3 +275,53 @@ class FileSystem(ABC, metaclass=ABCMeta):
             self.format_fs(format_options)
 
         self.mount(mountpoint, mount_options)
+
+    def freeze(self):
+        """
+        Freeze the filesystem
+
+        returns the output of the freeze command (optional).
+        """
+        try:
+            return run(
+                f"fsfreeze --freeze {self.mountpoint}",
+                check=True,
+                capture_output=True,
+            ).stdout.decode()
+        except subprocess.CalledProcessError as e:
+            if "Device or resource busy" in e.stderr.decode().strip():
+                logger.warning(
+                    "Filesystem already frozen.",
+                    filesystem=self.__filesystem__,
+                    mountpoint=self.mountpoint,
+                    device=self.device,
+                )
+                return
+            raise FileSystemFreezeError.from_exc(e, self.__filesystem__)
+        except Exception as e:
+            raise FileSystemFreezeError.from_exc(e, self.__filesystem__)
+
+    def unfreeze(self):
+        """
+        Unfreeze the filesystem
+
+        returns the output of the unfreeze command (optional).
+        """
+        try:
+            return run(
+                f"fsfreeze --unfreeze {self.mountpoint}",
+                check=True,
+                capture_output=True,
+            ).stdout.decode()
+        except subprocess.CalledProcessError as e:
+            if "Invalid argument" in e.stderr.decode().strip():
+                logger.warning(
+                    "Filesystem is not freezed.",
+                    filesystem=self.__filesystem__,
+                    mountpoint=self.mountpoint,
+                    device=self.device,
+                )
+                return
+            raise FileSystemUnFreezeError.from_exc(e, self.__filesystem__)
+        except Exception as e:
+            raise FileSystemUnFreezeError.from_exc(e, self.__filesystem__)
--- a/rawfile/filesystem/btrfs.py
+++ b/rawfile/filesystem/btrfs.py
@@ -2,7 +2,6 @@ from .base import (
     FileSystem as FileSystemBase,
     FileSystemFormatError,
     FileSystemResizeError,
-    FileSystemCreateSnapshotError,
     FileSystemDeleteSnapshotError,
 )
 from tempfile import TemporaryDirectory
@@ -14,6 +13,39 @@ class BTRFS(FileSystemBase):
     @property
     def __filesystem__(self) -> str:
         return "btrfs"
+
+    def delete_snapshot(self, name: str):
+        try:
+            with TemporaryDirectory(
+                prefix=f"rawfile.{self.__filesystem__}.fs-snapshot."
+            ) as tmp_dir:
+                try:
+                    snapshots_dir = f"{tmp_dir}/.snapshots"
+                    snapshot_subvol = f"{snapshots_dir}/{name}"
+                    run(
+                        f"""
+                        set -exo pipefail
+                        mount -t btrfs -o subvolid=0 {self.device} {tmp_dir}
+                        default_subvol_id=$(btrfs subvolume get-default "{tmp_dir}" | awk '{{print $2}}')
+                        default_subvol_path=$(btrfs subvolume list -o "{tmp_dir}" | awk -v id=$default_subvol_id '$2 == id {{print $NF}}')
+                        default_subvol="{tmp_dir}/$default_subvol_path"
+                        if [ -d "{snapshot_subvol}" ] || [ -f "{snapshot_subvol}" ]; then
+                          btrfs subvolume delete "{snapshot_subvol}"
+                        fi
+                        """,
+                        check=True,
+                        capture_output=True,
+                        executable="bash",
+                    )
+                finally:
+                    if get_device_for_mountpoint(tmp_dir):
+                        run(
+                            f"umount {tmp_dir}",
+                            check=True,
+                            capture_output=True,
+                        )
+        except Exception as e:
+            raise FileSystemDeleteSnapshotError.from_exc(e, self.__filesystem__)
 
     def format_fs(self, options: list[str] = []):
         with TemporaryDirectory(
@@ -55,68 +87,3 @@ class BTRFS(FileSystemBase):
             return output.stdout.decode()
         except Exception as e:
             raise FileSystemResizeError.from_exc(e, self.__filesystem__)
-
-    def create_snapshot(self, name: str):
-        try:
-            with TemporaryDirectory(
-                prefix=f"rawfile.{self.__filesystem__}.fs-snapshot."
-            ) as tmp_dir:
-                try:
-                    snapshots_dir = f"{tmp_dir}/.snapshots"
-                    snapshot_subvol = f"{snapshots_dir}/{name}"
-                    run(
-                        f"""
-                        set -exo pipefail
-                        mount -t btrfs -o subvolid=0 {self.device} {tmp_dir}
-                        default_subvol_id=$(btrfs subvolume get-default "{tmp_dir}" | awk '{{print $2}}')
-                        default_subvol_path=$(btrfs subvolume show "{tmp_dir}" | head -1 | awk '{{print $1}}')
-                        default_subvol="{tmp_dir}/$default_subvol_path"
-                        mkdir -p "{snapshots_dir}"
-                        btrfs subvolume snapshot -r "{tmp_dir}" "{snapshot_subvol}"
-                        """,
-                        check=True,
-                        capture_output=True,
-                        executable="bash",
-                    )
-                finally:
-                    if get_device_for_mountpoint(tmp_dir):
-                        run(
-                            f"umount {tmp_dir}",
-                            check=True,
-                            capture_output=True,
-                        )
-        except Exception as e:
-            raise FileSystemCreateSnapshotError.from_exc(e, self.__filesystem__)
-
-    def delete_snapshot(self, name: str):
-        try:
-            with TemporaryDirectory(
-                prefix=f"rawfile.{self.__filesystem__}.fs-snapshot."
-            ) as tmp_dir:
-                try:
-                    snapshots_dir = f"{tmp_dir}/.snapshots"
-                    snapshot_subvol = f"{snapshots_dir}/{name}"
-                    run(
-                        f"""
-                        set -exo pipefail
-                        mount -t btrfs -o subvolid=0 {self.device} {tmp_dir}
-                        default_subvol_id=$(btrfs subvolume get-default "{tmp_dir}" | awk '{{print $2}}')
-                        default_subvol_path=$(btrfs subvolume list -o "{tmp_dir}" | awk -v id=$default_subvol_id '$2 == id {{print $NF}}')
-                        default_subvol="{tmp_dir}/$default_subvol_path"
-                        if [ -f "{snapshot_subvol}" ]; then
-                          btrfs subvolume delete "{snapshot_subvol}"
-                        fi
-                        """,
-                        check=True,
-                        capture_output=True,
-                        executable="bash",
-                    )
-                finally:
-                    if get_device_for_mountpoint(tmp_dir):
-                        run(
-                            f"umount {tmp_dir}",
-                            check=True,
-                            capture_output=True,
-                        )
-        except Exception as e:
-            raise FileSystemDeleteSnapshotError.from_exc(e, self.__filesystem__)
--- a/rawfile/internal_svc.py
+++ b/rawfile/internal_svc.py
@@ -1,6 +1,15 @@
+import time
 from internal import internal_pb2_grpc, internal_pb2
 from utils.logs import logger, log_grpc_request
-import utils.rawfile
+from utils.rawfile import (
+    metadata,
+    truncate,
+    fallocate,
+    img_file,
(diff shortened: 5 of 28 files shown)


rawfile/internal_svc.py
```diff
@@ -42,30 +48,42 @@ def ExpandRawFile(
...
+            patch_metadata(
                 request.volume_id,
                 {"size": request.new_size},
             )
             return internal_pb2.ExpandRawFileResponse(
                 is_attached=is_attached(request.volume_id),
                 status=internal_pb2.ExpandRawFileStatus.OK,
             )
+
+    def GetRawFile(self, request: internal_pb2.GetRawFileRequest, context):
```
mhkarimi1383, line 79: It will use a Temporary snapshot later on...

rawfile/bd2fs.py
```diff
@@ -285,20 +283,37 @@ def ControllerExpandVolume(self, request, context):
...
+            if loop_devs and len(loop_devs):
+                for dev in loop_devs:
+                    mp = device_to_mountpoint(dev, first_only=False)
+                    if mp:
+                        mountpoints.extend(mp)
+            os.makedirs(snapshots_path, exist_ok=True)
+            if volume_meta.get("freezefs", False) and mountpoints:
+                for mp in mountpoints:
+                    run(f"fsfreeze --freeze {mp}", check=True)
+            cmd = f"cp --sparse=always --reflink={'always' if volume_meta.get('copy_on_write', False) else 'never'}"
```
laurivosandi: fsfreeze without reflinking would result in data corruption... convention over configuration...

mhkarimi1383: fs freeze only blocks write operations,

But if that's the case can you send some documents/test scenerios for that?
We can make this default and also warn users about that

But what about backing storage does not supper COW?

laurivosandi: 1. Reflinking should work without freezefs (XFS, Btrfs)
2. Freezefs could be used as fallback for other filesystems (ext4)
3. Anything else should explicitly fail. Why warn user if it will never work?

mhkarimi1383: Yeah,
We have decided to make these completely configurable

And make default options to be safe in every situation

But we will have a complete documentation on this and how options are working together

tiagolobocastro: without fsfreeze or reflink we should reject snapshot if volume is in use

mhkarimi1383: Yeah 👍

rawfile/rawfile_servicer.py
```diff
@@ -208,14 +223,19 @@ def CreateVolume(self, request, context):
                 grpc.StatusCode.INVALID_ARGUMENT, "Topology key not found... why?"
             )
 
-        params = normalize_parameters(request.parameters)
-        thin_provision = params.get("thinprovision", "no")
         format_options = params.get("formatoptions", "").strip()
+        snap_inuse = str_to_bool(params.get("snapinuse", "no"))
+        freezefs = str_to_bool(params.get("freezefs", "no"))
+        copy_on_write = str_to_bool(params.get("copyonwrite", "no"))
```
tiagolobocastro: shouldn't this come from filesystem where data is living on?

rawfile/rawfile_servicer.py
```diff
@@ -208,14 +223,19 @@ def CreateVolume(self, request, context):
                 grpc.StatusCode.INVALID_ARGUMENT, "Topology key not found... why?"
             )
 
-        params = normalize_parameters(request.parameters)
-        thin_provision = params.get("thinprovision", "no")
         format_options = params.get("formatoptions", "").strip()
+        snap_inuse = str_to_bool(params.get("snapinuse", "no"))
+        freezefs = str_to_bool(params.get("freezefs", "no"))
```
tiagolobocastro, line 236: this should come from the snapshot class rather than volume?

rawfile/rawfile_servicer.py
```diff
@@ -208,14 +223,19 @@ def CreateVolume(self, request, context):
                 grpc.StatusCode.INVALID_ARGUMENT, "Topology key not found... why?"
             )
 
-        params = normalize_parameters(request.parameters)
-        thin_provision = params.get("thinprovision", "no")
         format_options = params.get("formatoptions", "").strip()
+        snap_inuse = str_to_bool(params.get("snapinuse", "no"))
```
tiagolobocastro: what does this mean?

mhkarimi1383: This option controls if we are able to take snapshot from an in use volume or not

rawfile/rawfile_servicer.py
mhkarimi1383: You mean we should determine them from the data dir filesystem?

rawfile/rawfile_servicer.py
mhkarimi1383: I have used storage class parameters because we will need them outside of the snapshots

rawfile/rawfile_servicer.py
mhkarimi1383: We can also set this automatically based on COW and fsfreeze

rawfile/rawfile_servicer.py
tiagolobocastro: Yes

rawfile/rawfile_servicer.py
tiagolobocastro: Ah ok, where else would it be used then?

rawfile/rawfile_servicer.py
tiagolobocastro: yes :)

mhkarimi1383: done.

rawfile/rawfile_servicer.py
mhkarimi1383: for temporary snapshots for between node migration/clones
that's why I have set them in volume metadata

rawfile/bd2fs.py
mhkarimi1383: done. it is getting rejected right now

mhkarimi1383:
I have also fixed size problem, it is not `0` anymore, and I have added `ListSnapshots` too and if snapshot size changes it will get reported

rawfile/rawfile_servicer.py
mhkarimi1383: I will check if we can determine that from underlining filesystem of data dir, So we can set copy on write automatically

mhkarimi1383:
@tiagolobocastro 
With this we can do snapshots in background I think :)

So if snapshot creation takes longer we won't get timeout errors

rawfile/rawfile_servicer.py
mhkarimi1383: There is not way to check that without checking out FS and it's mount info

in xfs for example

```console
xfs_info /
meta-data=/dev/disk/by-uuid/72fcf69a-04b3-481a-9a94-2164a4c2d2b5 isize=512    agcount=4, agsize=62381653 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=1
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=1
         =                       exchange=0   metadir=0
data     =                       bsize=4096   blocks=249526609, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1, parent=0
log      =internal log           bsize=4096   blocks=121839, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
         =                       rgcount=0    rgsize=0 extents
         =                       zoned=0      start=0 reserved=0
```

`reflink=1` means I have COW, and in others we should check them by their FS too, or we should test reflinking before creating snapshots

mhkarimi1383:
@mergifyio rebase

rawfile/rawfile_servicer.py
mhkarimi1383: I want to also add storage pools later on, So we will have these params there

rawfile/rawfile_servicer.py
tiagolobocastro: So we could do this then?

mhkarimi1383: We can, but it is so dirty also users should be able to manage this with full control

rawfile/bd2fs.py
```diff
@@ -285,24 +285,49 @@ def ControllerExpandVolume(self, request, context):
...
+            if (loop_devs and len(loop_devs)) and (not snap_inuse):
+                context.abort(
+                    grpc.StatusCode.FAILED_PRECONDITION,
+                    "Taking a snapshot of a mounted volume requires either COW or FreezeFS to be enabled.",
                 )
-            fs.create_snapshot(name=request.name)
-
-            if fs.mountpoint is None:
-                detach_loops(file)
+                return
```
tiagolobocastro: pardon my ignorance, does this return, return the error?

rawfile/bd2fs.py
```diff
@@ -285,24 +285,49 @@ def ControllerExpandVolume(self, request, context):
...
+                for dev in loop_devs:
+                    mp = device_to_mountpoint(dev, first_only=False)
+                    if mp:
+                        mountpoints.extend(mp)
+            should_freeze = all((freezefs, mountpoints))
+            try:
+                os.makedirs(snapshots_path, exist_ok=True)
+                if should_freeze:
+                    for mp in mountpoints:
+                        run(f"fsfreeze --freeze {mp}", check=True)
```
tiagolobocastro: what happens if the pod is deleted whilst freezed, who does the unfreeze?
or I guess we always wait for create_snapshot to conclude on retry?

rawfile/bd2fs.py
```diff
@@ -285,24 +285,49 @@ def ControllerExpandVolume(self, request, context):
...
+                Path(f"{snapshots_path}/{request.name}.img.creating").touch()
+                cmd = f"cp --sparse=always --reflink={'always' if copy_on_write else 'never'}"
```
tiagolobocastro: presumably this will take up a long time and cause some issues with csi retrying?
Should have some kind of "copy manager" which could detect an existing copy is in-progress and return an appropriate error?

tiagolobocastro: Also there's no progress report this way, can anything be done like for example, add annotation every X GiB copied?

rawfile/protos/internal.proto
```diff
@@ -1,7 +1,10 @@
 syntax = "proto3";
 
+import "google/protobuf/any.proto";
+
 service Internal {
   rpc ExpandRawFile (ExpandRawFileRequest) returns (ExpandRawFileResponse);
+  rpc GetRawFile (GetRawFileRequest) returns (GetRawFileResponse);
```
tiagolobocastro, line 9: Must have missed it, when is this used?

mhkarimi1383: Clone and migrate between nodes

rawfile/rawfile_servicer.py
tiagolobocastro: This may become very difficult to manage though as this is a global value
I have seen clusters start very homogeneously, same nodes, memory, storage etc and as things evolve then different nodes get added, different kernel version, different OS, etc so it's very possible that not all nodes have the same COW filesystem?

rawfile/protos/internal.proto
tiagolobocastro: Is that on this PR already or is this just adding the infra for it?

mhkarimi1383: Will be implemented here

rawfile/bd2fs.py
mhkarimi1383: Yeah we should somehow track it

rawfile/bd2fs.py
mhkarimi1383: We should make unmounts to use lock, so it will wait

rawfile/bd2fs.py
mhkarimi1383: Yeah it is after context.abort

mhkarimi1383:
@mergify rebase

mhkarimi1383:
> @tiagolobocastro With this we can do snapshots in background I think :)
> 
> So if snapshot creation takes longer we won't get timeout errors

@tiagolobocastro What do you think about this

I can add some background workers for handling this using threads

tiagolobocastro:
> > @tiagolobocastro With this we can do snapshots in background I think :)
> > So if snapshot creation takes longer we won't get timeout errors
> 
> @tiagolobocastro What do you think about this
> 
> I can add some background workers for handling this using threads

Yeah possibly, with some way of querying if existing task already in progress? Is there a csi error for this scenario or is it expected to just set a huge timeout value?

Btw I suggest you split this into PRs. Ex start with the reflink only as we don't need to worry about long copies with that right?

mhkarimi1383:
> > > @tiagolobocastro With this we can do snapshots in background I think :)
> > > So if snapshot creation takes longer we won't get timeout errors
> > 
> > 
> > @tiagolobocastro What do you think about this
> > I can add some background workers for handling this using threads
> 
> Yeah possibly, with some way of querying if existing task already in progress? Is there a csi error for this scenario or is it expected to just set a huge timeout value?
> 
> Btw I suggest you split this into PRs. Ex start with the reflink only as we don't need to worry about long copies with that right?

I will try to split PRs I will make it to run in background and track progress somehow

rawfile/rawfile_servicer.py
mhkarimi1383: This is like what we have for LVM and others that we are settings a storage pool with some features enabled or disabled

If we add storage pool support we can get these params per pool and per node

mhkarimi1383:
I'm working on a snapshot manager that could handle graceful shutdown

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,98 @@
...
+                creation_time = time.time()
+                return Snapshot(
+                    name=name,
+                    volume_id=volume_id,
+                    size_bytes=snap_path.stat().st_size,
+                    creation_time=creation_time,
+                )
+
+            finally:
+                Path(f"{snap_path}.creating").unlink(missing_ok=True)
```
mhkarimi1383: We need to also make to not mark failed ones as read to use

rawfile/utils/task_manager.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class TaskInfo(TypedDict):
+    task: str
+    args: list
+    kwargs: dict
+    retry_count: int
+
+
+class TaskManager:
```
mhkarimi1383, line 44: A global shared mux lock is required and some duplicate codes should be removed

mhkarimi1383:
Something is strange is that external snapshotter is not calling `ListSnapshot` and snapshots are not getting ready

mhkarimi1383:
Also `ControllerGetCapabilities` response is empty


```
2025-09-25 09:12:33.194 | INFO     | utils.logs:wrap:130 - GRPC Server Access Log | {'handler': 'ControllerGetCapabilities', 'request': , 'starttime': datetime.datetime(2025, 9, 25, 9, 12, 33, 194573), 'response': , 'latency': datetime.timedelta(microseconds=48), 'endtime': datetime.datetime(2025, 9, 25, 9, 12, 33, 194621), 'success': True}
```

mhkarimi1383:
@mergify rebase

mhkarimi1383:
We don't have progress percentage tracking, But we have a live update of the size and readiness of the snapshot creation also create time is a real value right now 

And we have a graceful shutdown for that

Also we can move some other parts to use the task manager

mhkarimi1383:
I have to also change the snapshot list and remove to use the new manager

Then we should add a volume manager too I think 🤔 
Since restore of large snapshots may take longer than gRPC timeout

rawfile/bd2fs.py
```diff
@@ -284,22 +289,36 @@ def ControllerExpandVolume(self, request, context):
...
+            if current:
+                snapshot = current[0]
+                return csi_pb2.CreateSnapshotResponse(
+                    snapshot=csi_pb2.Snapshot(
+                        size_bytes=snapshot.size_bytes,
+                        snapshot_id=f"{snapshot.volume_id}/{snapshot.name}",
+                        source_volume_id=snapshot.volume_id,
+                        ready_to_use=snapshot.ready,
+                        creation_time=Timestamp(
+                            seconds=int(snapshot.creation_time),
```
tiagolobocastro: this is implemented differently below

rawfile/bd2fs.py
```diff
@@ -308,7 +327,7 @@ def CreateSnapshot(self, request: csi_pb2.CreateSnapshotRequest, context):
                     creation_time=Timestamp(
                         seconds=creation_time_ns // nano, nanos=creation_time_ns % nano
                     ),
-                    ready_to_use=True,
+                    ready_to_use=False,
```
tiagolobocastro: If we're using copy_on_write isn't the snapshot instance, in which can we can return snapshot as ready?

rawfile/rawfile_servicer.py
tiagolobocastro: We could make it configurable but I think by default we should figure out the correct value for cow based on the data filesystem; don't need to do it all at once though

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,115 @@
...
+        name: str,
+        copy_on_write: bool = False,
+        freeze_fs: bool = False,
+    ) -> Snapshot:
+        with VolLock(volume_id):
+            file = img_file(volume_id)
+            loop_devs = attached_loops(file.as_posix())
+            snap_path = self._get_snapshot_path(volume_id, name)
+            snap_path.parent.mkdir(parents=True, exist_ok=True)
+            Path(f"{snap_path}.creating").touch()
```
tiagolobocastro, line 52: Should we add metadata files for snapshots as well rather than just looking at img files?

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,115 @@
...
+                    mp = device_to_mountpoint(dev, first_only=False)
+                    if mp:
+                        mountpoints.extend(mp)
+            try:
+                if freeze_fs and mountpoints:
+                    for mp in mountpoints:
+                        run(f"fsfreeze --freeze {mp}", check=True)
+
+                reflink = "always" if copy_on_write else "never"
+                cmd = f"cp --sparse=always --reflink={reflink} {file} {snap_path}"
```
tiagolobocastro: btw seems cp returns not supported with reflink=always if the fs doesn't support it, could we make use of this to determine cow?

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,115 @@
...
+                    size_bytes=snap_path.stat().st_size,
+                    creation_time=creation_time,
+                )
+            except Exception as e:
+                Path(snap_path).unlink(missing_ok=True)
+                raise e
+            finally:
+                if freeze_fs and mountpoints:
+                    for mp in mountpoints:
+                        run(f"fsfreeze --unfreeze {mp}", check=True)
```
tiagolobocastro: in case of a crash who unfreezes, is it the task manager by retrying the create again?
and if so, what happens if snapshot is deleted meanwhile?

rawfile/utils/snapshot_manager.py
mhkarimi1383: If that's the case we can do a fake cp command to determine that on startup, and save it into memory

rawfile/bd2fs.py
mhkarimi1383: This is for handling snapshot creation in background, for progress tracking external snapshotter sends multiple snapshot create multiple times, I will check that if the snapshot already exists I will return the existing one's information

rawfile/bd2fs.py
mhkarimi1383: We are delighting the entire process to task manager and we know if the worker has picked up the job or not

rawfile/utils/snapshot_manager.py
mhkarimi1383: We have graceful shutdown, but we could add an static job on startup to unfreeze everything before starting the task manager

rawfile/utils/snapshot_manager.py
mhkarimi1383: By metadata files, we will look at img files later on too, since we need the exact size of the file

rawfile/bd2fs.py
tiagolobocastro: I had meant the creation_time only, seems to be set differently though perhaps because we're taking i from different sources?

rawfile/bd2fs.py
tiagolobocastro: Sorry didn't get you?

rawfile/utils/snapshot_manager.py
tiagolobocastro: yes that would be neat :)

rawfile/bd2fs.py
mhkarimi1383: Yeah, initially it will be fake, but later on it will get changed to the actual ctime of the file

rawfile/utils/snapshot_manager.py
tiagolobocastro: And also what happens if we receive another csi call whilst we have a copy manager?
Ex: 
- snapshot may be taken whilst volume is unpublished, then we copy... and then we receive a controllerPublish
- snapshot is taken whilst volume is staged on a node, then we receive a nodePublish or nodeUnpublish
- etc

rawfile/bd2fs.py
mhkarimi1383: Even using CoW, we are doing it in the background using the background worker, So we don't know if the task manager has picked up the job or not

tiagolobocastro: Yeah that's my doubt, do we need copy manager if we're using cow? 
Otherwise we could have a small wait with timeout in case it's cow?

rawfile/utils/snapshot_manager.py
mhkarimi1383: I will make it to determine that by default, but users are able to override that

rawfile/utils/snapshot_manager.py
mhkarimi1383: We are using volume locks in all of them So other Operations are blocked when doing snapshot

rawfile/bd2fs.py
mhkarimi1383: Okay, we can wait for 30s for that in the foreground

mhkarimi1383: Done.

rawfile/utils/snapshot_manager.py
mhkarimi1383: So I think it is not necessary

rawfile/rawfile_servicer.py
mhkarimi1383: I will handle this :)

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,141 @@
...
+                raise SnapshotCreateVolumeInUse(volume_id=volume_id, snapshot_name=name)
+            if loop_devs and len(loop_devs):
+                for dev in loop_devs:
+                    mp = device_to_mountpoint(dev, first_only=False)
+                    if mp:
+                        mountpoints.extend(mp)
+            try:
+                if freeze_fs and mountpoints:
+                    for mp in mountpoints:
+                        run(f"fsfreeze --freeze {mp}", check=True)
```
mhkarimi1383: This should be added to filesystem interface since it is different in some filesystems (like xfs)

rawfile/utils/snapshot_manager.py
mhkarimi1383: Done.

rawfile/rawfile_servicer.py
mhkarimi1383: Done.

rawfile/utils/snapshot_manager.py
tiagolobocastro: What if the pod restarts?

mhkarimi1383: a static unfreeze on startup should fix that, that's the only idea that I have

mhkarimi1383:
@mergify rebase

rawfile/utils/snapshot_manager.py
mhkarimi1383: Why doing FSFreeze when we can Set entire blockdev to be RO?
It is easier to manage and also `attach_loop` will set them to RW automatically

mhkarimi1383:
I want to deprecate and remove filesystem level snapshots :)

rawfile/utils/snapshot_manager.py
mhkarimi1383: @tiagolobocastro the only this that is left is this one :)

rawfile/utils/remote.py
```diff
@@ -48,21 +49,37 @@ def init_rawfile(volume_id, size, thin_provision=False):
         img_file = Path(f"{img_dir}/disk.img")
         if img_file.exists() and os.path.getsize(img_file) >= size:
             return
+        snapshots_dir = Path(img_dir.joinpath("snapshots"))
+        os.makedirs(name=snapshots_dir, exist_ok=True)
         utils.rawfile.patch_metadata(
             volume_id,
             {
                 "schema_version": LATEST_SCHEMA_VERSION,
```
tiagolobocastro, line 57: Do we need to bump the version?

rawfile/protos/internal.proto
tiagolobocastro: Could you add some comments here explaining what it is for?

rawfile/utils/rawfile.py
```diff
@@ -65,6 +69,12 @@ def destroy(volume_id, dry_run=True):
         Path(img_file(volume_id)).unlink(missing_ok=True)
         Path(meta_file(volume_id)).unlink(missing_ok=True)
         Path(lock_file(volume_id)).unlink(missing_ok=True)
+        for each in snapshots_dir(volume_id).glob("*"):
```
tiagolobocastro: Deleting a volume should not invalidate the snapshots
In case of cow I guess we'll have to "fake" delete the volume, and in case of full copy it shouldn't matter?

rawfile/utils/remote.py
```diff
@@ -48,21 +49,37 @@ def init_rawfile(volume_id, size, thin_provision=False):
...
                 "thin_provision": thin_provision,
+                "freezefs": freezefs,
+                "copy_on_write": copy_on_write,
             },
         )
+        img_file.touch()
+        if snapshot_id:
+            volume_id, name = snapshot_id.rsplit("/", 1)
+            metadata = utils.rawfile.metadata(volume_id)
+            size = max(size, metadata["size"])
```
tiagolobocastro, line 72: Shouldn't the size match exactly the snapshot size?

rawfile/utils/snapshot_manager.py
tiagolobocastro: > Why doing FSFreeze when we can Set entire blockdev to be RO?

Not sure if that works, IIRC there were some pitfalls with setting the device as RO
Also we need to ensure the data if flushed to disk before we can start copying

> a static unfreeze on startup should fix that, that's the only idea that I have

That needs coordination with copy, as if we republish then we have to start copying snapshot back from scratch

rawfile/rawfile_servicer.py
tiagolobocastro: Should freezefs be turned on by default?

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,152 @@
...
+    def restore_snapshot(self, volume_id: str, name: str, destination: Path):
+        """Restore a snapshot"""
+        chunk_size = 1024 * 1024
+        snap_path = self._get_snapshot_path(volume_id, name)
+        with open(snap_path, "rb") as src, open(destination, "wb") as dst:
+            while True:
+                buf = src.read(chunk_size)
+                if not buf:
+                    break
+                dst.write(buf)
```
tiagolobocastro, line 119: Could we also make use of cow here?

rawfile/utils/remote.py
mhkarimi1383: No, because we are falling back to default when getting new keys

rawfile/utils/remote.py
mhkarimi1383: Idk, I have made that to the maximum size between snapshot and volume request

rawfile/utils/snapshot_manager.py
mhkarimi1383: Using COW will make the newly created volume depend on the source snapshot

rawfile/utils/rawfile.py
mhkarimi1383: Or we can simply reject volume deletion when we have snapshots I think some other local PVs are doing the same

rawfile/utils/remote.py
tiagolobocastro: For example let's say it's a filesystem volume, if the size is greater than volume we'd have to expand the filesystem? Which could be done though might be easier to ensure it's the same for now and expand later

mhkarimi1383: Currently it will expand it.
I'll make to to not expand if that's preferred

rawfile/utils/snapshot_manager.py
tiagolobocastro: Yeah, this may be ok as a cheap way of checking snapshot contents for example
Ie if copy_on_write property in the SC is true, then we use cow here, otherwise it's a full copy

rawfile/utils/rawfile.py
tiagolobocastro: At least zfs and mayastor succeed the csi call but keep the volume data around

rawfile/utils/remote.py
tiagolobocastro: Oh it's already expanding the filesystem? Awesome :D 
Then that's fine :+1:

rawfile/utils/snapshot_manager.py
mhkarimi1383: What if user deletes the volume/snapshot that is used as source

rawfile/utils/remote.py
mhkarimi1383: Yeah :)

rawfile/utils/rawfile.py
mhkarimi1383: I think LVM rejects it,

But we will need GC worker to check for them and remove them

rawfile/utils/snapshot_manager.py
mhkarimi1383: We are already doing it

Task manager will retry

(Only on graceful shutdowns)
I will add a checker for snapshots to retry stuck ones on startup

rawfile/utils/rawfile.py
tiagolobocastro: We might need to change LVM then, this is also a change we made on ZFS IIRC.
Could we delete it as part of snapshot delete? Ex: when last reference is deleted we delete the volume

rawfile/utils/snapshot_manager.py
tiagolobocastro: We would have to keep the data around and delete it when last reference is deleted; this is how we do it on mayastor
But we could do it as subsequent change to keep it simpler here for now

rawfile/rawfile_servicer.py
mhkarimi1383: Default behavior should not block running applications

rawfile/utils/snapshot_manager.py
mhkarimi1383: OK, For that I have to change the GC behavior and make it to also run after snapshot deletion

rawfile/rawfile_servicer.py
tiagolobocastro: Shouldn't default behavior be safety first, a consistent snapshot? Other user may have nasty surprise later when trying to use it
So I think `dangerous` should be an opt-in

rawfile/utils/rawfile.py
mhkarimi1383: I will make Delete Snapshot to do GC too and check for this condition

rawfile/rawfile_servicer.py
mhkarimi1383: We are already blocking attached snapshots

Both are OK if we document that

tiagolobocastro: doh, forgot that we were already blocking `createSnapshot`, my bad!
Then I think we're good with that
We should also start documenting these new changes

rawfile/utils/remote.py
tiagolobocastro: Alright

rawfile/rawfile_servicer.py
mhkarimi1383: Yeah

any updates on product document itself?

rawfile/utils/snapshot_manager.py
mhkarimi1383: I have handled crashes during snapshot creation

I think it is better to handle all of the stuck tasks, So we will have safe clones later on (I will refactor) :)

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify rebase

rawfile/filesystem/base.py
```diff
@@ -272,3 +288,33 @@ def format_and_mount(
...
+
+    def freeze(self):
+        """
+        Freeze the filesystem
+
+        returns the output of the freeze command (optional).
+        """
+        try:
+            run(
+                f"fsfreeze --freeze {self.mountpoint}",
```
tiagolobocastro, line 287: Missing idempotency handling? IRRC errors are return if fs is already frozen/unfrozen

rawfile/protos/internal.proto
```diff
@@ -1,20 +1,46 @@
 syntax = "proto3";
 
+import "google/protobuf/any.proto";
+
 service Internal {
+  // Expands raw file to new size.
   rpc ExpandRawFile (ExpandRawFileRequest) returns (ExpandRawFileResponse);
+  // Downlaods raw file data and metadata.
```
tiagolobocastro:
```suggestion
  // Downloads raw file data and metadata.
```

rawfile/protos/internal.proto
```diff
@@ -1,20 +1,46 @@
...
+  string volume_id = 1;
+  // Whether to include file data in the response.
+  bool with_data = 2;
+  // Optional snapshot ID to get the raw file at a specific snapshot. will create a temporary snapshot if not provided.
+  optional string snapshot_id = 3;
+}
+message GetRawFileResponse {
+  // Metadata associated with the raw file.
+  map<string, google.protobuf.Any> metadata = 1;
+  // Raw file data. Present only if with_data was true in the request.
```
tiagolobocastro: nit: throughout should "raw file" be a single work? "rawfile"

```suggestion
  // Rawfile data. Present only if with_data was true in the request.
```

rawfile/protos/internal.proto
```diff
@@ -1,20 +1,46 @@
...
+  // Status of the expand operation.
   ExpandRawFileStatus status = 2;
 }
+
+message GetRawFileRequest {
+  // Volume ID of the raw file to be downloaded.
+  string volume_id = 1;
+  // Whether to include file data in the response.
+  bool with_data = 2;
+  // Optional snapshot ID to get the raw file at a specific snapshot. will create a temporary snapshot if not provided.
```
tiagolobocastro, line 38: how do we keep track and gc this temporary snapshot?

rawfile/utils/rawfile.py
```diff
@@ -55,16 +55,27 @@ def img_file(volume_id):
...
+
 def img_size(volume_id) -> int:
     return metadata(volume_id)["size"]
 
 
 def destroy(volume_id, dry_run=True):
     logger.info("Destroying Volume", volume_id=volume_id, dry_run=dry_run)
     if not dry_run:
+        snapshots = list(snapshots_dir(volume_id).glob("*"))
+        if len(snapshots) > 0:
```
tiagolobocastro: Might be worth adding a log here saying something like "Forgetting about volume xxx because it has yyy dependent snapshots" ?

tiagolobocastro: Ah also, should we check if the snapshots are full copy? If they're full copy then we can delete the volume?

rawfile/utils/task_manager.py
```diff
@@ -0,0 +1,195 @@
...
+            logger.opt(exception=exc).error(
+                "Task failed", task_id=task_id, task=task_info, exception=exc
+            )
+        task_info["state"] = state
+        self.save_task(task_id, task_info)
+
+    def save_task(self, task_id: str, task_info: TaskInfo):
+        with self._lock:
+            logger.debug("Saving task", task_id=task_id, task=task_info)
+            with open(self._tasks_store_path, "r+") as tasks_file:
```
tiagolobocastro, line 153: would it be safer to have unique files per task rather than a single for all?

rawfile/utils/task_manager.py
```diff
@@ -0,0 +1,195 @@
...
+                current: dict[str, TaskInfo] = json.loads(tasks_file.read() or "{}")
+                current[task_id] = task_info
+                tasks_file.truncate(0)
+                tasks_file.seek(0)
+                json.dump(current, tasks_file)
```
tiagolobocastro, line 158: fsync as well?

docs/design/taskmanager.md
```diff
@@ -0,0 +1,96 @@
...
+## Technical Implementation
+
+### 1. Thread Safety
+
+- Uses `threading.Lock` for file operations
+- Ensures atomic operations on task storage
+
+### 2. Task Lifecycle
+
+1. Task submission
```
tiagolobocastro, line 68: Should we also add a max number of tasks? IE with a max of 16 tasks, if we get another create_snapshot or create_clone then we'd reject it

rawfile/rawfile.py
```diff
@@ -34,15 +38,36 @@ def csi_driver(driver_config: CSIDriverCmd):
...
-        bd2fs.Bd2FsIdentityServicer(rawfile_servicer.RawFileIdentityServicer()), server
+        bd2fs.Bd2FsIdentityServicer(
+            rawfile_servicer.RawFileIdentityServicer(),
+            task_manager=_task_manager,
+        ),
+        server,
     )
     if driver_config.plugin_type == "node":
         migrate_all_volume_schemas()
+        cow_test_src = Path(f"{consts.DATA_DIR}/cow.test.src")
```
tiagolobocastro: maybe move this logic to the filesystem/init.py ?

docs/design/taskmanager.md
```diff
@@ -0,0 +1,96 @@
...
+
+- Graceful shutdown with timeout
+- Marks running tasks as failed
+- Cancels pending futures
+- Preserves task states
+
+## Error Handling
+
+- Custom `TaskManagerShuttingDown` exception
+- Automatic crash recovery
```
tiagolobocastro, line 83: How are tasks synchronized with csi operations after pod or node restart?

rawfile/bd2fs.py
```diff
@@ -284,55 +288,104 @@ def ControllerExpandVolume(self, request, context):
...
+                volume_id=request.source_volume_id, snapshot_name=request.name
+            )
+            if not _current.data:
+                return None
+            return _current.data[0]
+
+        nano = 10**9
+        current = _get_current_snapshot()
+        if not current:
+            self._task_manager.run_task(
```
tiagolobocastro, line 318: there's a race condition here during the time between letting go of the `VolLock` and re-acquiring it here on the `run_task::snapshot_manager::create_snapshot` ?

rawfile/utils/snapshot_manager.py
```diff
@@ -0,0 +1,156 @@
...
+class SnapshotList:
+    data: list[Snapshot]
+    next_token: int | None
+
+
+class SnapshotManager:
+    def _get_snapshot_path(self, volume_id: str, name: str) -> Path:
+        return snapshots_dir(volume_id) / f"{name}.img"
+
+    def create_snapshot(
```
tiagolobocastro, line 40: in case of pod/node restart do we always retry the tasks ahead of any csi operations? or is there a chance a csi operation could get through?

tiagolobocastro (commented):
Have you tested a large copy with cow off and restart the pod mid-way? I suspect it will fail as we're not handling fsfreeze idempotency

rawfile/utils/rawfile.py
mhkarimi1383: We can delete the img, meta and lock file
But what if the user didn't specify the COW and changed the filesystem, we don't know if that's COW or not

rawfile/utils/snapshot_manager.py
mhkarimi1383: There is a chance for that, we will just mark them as failed and retry worker will pick them

rawfile/bd2fs.py
mhkarimi1383: Here we only need lock while reading meta file, nothing will happen if we create multiple snapshots of a single volume at the same time

docs/design/taskmanager.md
mhkarimi1383: They are already in the task sore json file with their state

docs/design/taskmanager.md
mhkarimi1383: We have a limited thread pool, I will do a quick test to see what will happen, they will get queued I think

rawfile/protos/internal.proto
mhkarimi1383: I have just made an infra for that here, but It will be handled just like other snapshots

rawfile/utils/task_manager.py
mhkarimi1383: Handling a specific file per task will be complicated, I know we don't need locking of task file in some places anymore, but it will be complicated I think

rawfile/filesystem/base.py
mhkarimi1383: I will test it out and fix it

rawfile/rawfile.py
mhkarimi1383: Yeah, it should be moved to a better place

rawfile/utils/rawfile.py
tiagolobocastro: We could use a snapshot json configuration file?

rawfile/utils/snapshot_manager.py
tiagolobocastro: We probably need to ensure the CSI operations don't get in the way or we'd have to make sure we unfreeze the filesystem on mount, because the fs will still be frozen if you unmount/remount for example.

rawfile/utils/task_manager.py
tiagolobocastro: Yeah possibly, was just wondering about write atomicity, and what happens if the node crashes half-way through the write
Maybe just the fsync for now is fine

rawfile/utils/snapshot_manager.py
mhkarimi1383: We can make it unfreeze all on startup

or we can say that this is expected while snapshot creation is in progress

tiagolobocastro: If we don't unfreeze but accept the csi nodeStage call, we risk being frozen forever if not cow, because in that case snapshot task would fail with `SnapshotCreateVolumeInUse`?

mhkarimi1383: If we have enabled freeze it we not fail with SnapshotCreateVolumeInUse

tiagolobocastro: oh I see, I misread the code below
then the app would just get stuck until the snapshot is taken, I think this is fine then?

mhkarimi1383: Yep it is expected

rawfile/rawfile.py
mhkarimi1383: I will add some startup hooks, I will move some parts there

docs/design/taskmanager.md
tiagolobocastro: As per the other thread, I think the answer is they are not really synchronized beyond they both take the volume lock? As in, there's no specific log for example: csi_request(): if snap in progress do_this else do_that ?
I wonder if we could massage the returned error to csi by saying snapshot in progress or something?

tiagolobocastro (approved):
Only fsfreeze idempotency and the volume deletion thingy outstanding, otherwise this is looking good

docs/design/taskmanager.md
mhkarimi1383: Duplicate calls is already handled
we are hashing the name and params of the task and we are using it as task id
So it will not rerun duplicate call

Currently to track if snapshot has been done or not, external snapshotter will call CreateSnapshot multiple times

rawfile/utils/task_manager.py
mhkarimi1383: Yeah
I made it to do fsync and we have locks too

rawfile/rawfile.py
mhkarimi1383: Done.

rawfile/utils/rawfile.py
mhkarimi1383: Or we can save if we are using COW or not in the snapshot annotation itself

rawfile/filesystem/base.py
mhkarimi1383: I will ignore resource unavailable errors since there is no direct way to check

mhkarimi1383: Done.

mhkarimi1383:
I will also make it to reject block volumes since we can't do fsfreeze on them

mhkarimi1383:
@tiagolobocastro 

FsFreeze problems fixed

rawfile/filesystem/base.py
```diff
@@ -272,3 +289,53 @@ def format_and_mount(
...
+        returns the output of the freeze command (optional).
+        """
+        try:
+            return run(
+                f"fsfreeze --freeze {self.mountpoint}",
+                check=True,
+                capture_output=True,
+            ).stdout.decode()
+        except subprocess.CalledProcessError as e:
+            if "Device or resource busy" in e.stderr.decode().strip():
```
tiagolobocastro, line 292: we could also replace this run with an ioctl and then check the codes

rawfile/utils/rawfile.py
tiagolobocastro: Yes perhaps though might be tricky to get that information here
A config file or even a simple file marker would probably be enough

tiagolobocastro (approved):
Cool, once this gets in I'll give this a whack
Wonder if there's some other existing csi e2e tests we could leverage for e2e testing

rawfile/filesystem/base.py
mhkarimi1383: I have read the fsfreeze code itself, and fsfreeze has no return code related to this

mhkarimi1383:
> Cool, once this gets in I'll give this a whack
> Wonder if there's some other existing csi e2e tests we could leverage for e2e testing

I this we could enable e2e tests for snapshot and restores

mhkarimi1383:
Also we need snapshot classes in chart too

rawfile/utils/rawfile.py
mhkarimi1383: In volume meta I'll put that if this volume has reflinks attached to it to or not

mhkarimi1383: Done.

rawfile/filesystem/base.py
mhkarimi1383: the ioctl will return EBUSY if already frozen, according to https://www.man7.org/linux/man-pages/man3/errno.3.html it means Device or resource busy

mhkarimi1383:
I have enabled snapshot restore test in the E2E test

mhkarimi1383:
@tiagolobocastro 

It expect when we remove snapshot content provisener should create it or something like that.

Is that a thing that we should cover?

https://github.com/openebs/rawfile-localpv/actions/runs/18398401445/job/52422116518?pr=213#step:15:7665

This should be handled using the external snapshotter and external provisioner since we are managing snapshot content directly

rawfile/filesystem/base.py
tiagolobocastro: Yes, found the mayastor implementation: https://github.com/openebs/mayastor-control-plane/blob/bdb6a11f0733f0134e3db188b9354a7e2ca286db/control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs#L31

tiagolobocastro:
> @tiagolobocastro
> 
> It expect when we remove snapshot content provisener should create it or something like that.
> 
> Is that a thing that we should cover?
> 
> https://github.com/openebs/rawfile-localpv/actions/runs/18398401445/job/52422116518?pr=213#step:15:7665
> 
> This should be handled using the external snapshotter and external provisioner since we are managing snapshot content directly

No, we need not do anything, the side car should handle it. I tend to think of the content like PV for a PVC - it represents the storage itself but we don't have to deal with it directly.

rawfile/filesystem/base.py
mhkarimi1383: So, No need to use ioctl here I think

mhkarimi1383:
> > @tiagolobocastro
> > It expect when we remove snapshot content provisener should create it or something like that.
> > Is that a thing that we should cover?
> > https://github.com/openebs/rawfile-localpv/actions/runs/18398401445/job/52422116518?pr=213#step:15:7665
> > This should be handled using the external snapshotter and external provisioner since we are managing snapshot content directly
> 
> No, we need not do anything, the side car should handle it. I tend to think of the content like PV for a PVC - it represents the storage itself but we don't have to deal with it directly.

I have to disable that test I think

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@tiagolobocastro

I'm almost done with this one, I will make CreateVolume to support cloning and make CreateVolume to work in background in a separated PR then.

If everything is OK I will squash 

note that I have deprecated FileSystem level BTRfs snapshots, We should consider telling users that

```
FileSystem Level snapshots should be removed before upgrade to avoid
conflicts
```

rawfile/filesystem/btrfs.py
tiagolobocastro, line 1: Don't we need to leave the deletion around to allow deletion of any existing fs snapshots?

tiagolobocastro (approved):

rawfile/filesystem/btrfs.py
mhkarimi1383: Hmmm...

In `DeleteSnapshot` rpc we should handle that if snapshot does not exist in the block volumes it should be handled via filesystem level handler

mhkarimi1383:
Users should warn that we will only allow deletion of snapshots

and this functionality will be removed

mhkarimi1383:
@mergify squash

tiagolobocastro (approved):

mhkarimi1383:
@mergify queue

PR #217, opened 2025-09-13
:memo: Add k3s image load command for nixos-shell

mhkarimi1383:
Since K3s is using containerd it is required to load them manually after building image


tiagolobocastro (approved):
We could also modify the scripts to detect and do this

mhkarimi1383:
> We could also modify the scripts to detect and do this

Yeah but we will say that in document for now

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify queue

PR #226, opened 2025-10-03
ci: deployer with xfs and reflink

tiagolobocastro:
Deployer now creates the rawfile path with an xfs formatted rawfile, with reflink enabled which is then mounted as a loop device.

Almost like using rawfile for rawfile :-)


mhkarimi1383 (approved):

tiagolobocastro:
@mergify queue

tiagolobocastro:
@mergify queue

PR #228, opened 2025-10-04
Re-enable Block BDD tests

tiagolobocastro:
test(bdd/block): wait for node unstage on expansion test

    This allows the block test to proceed without NodeExpansion
    todo: since we are checking if volume is served why not do
    expansion directly?

    Also don't log the retrying methods every time as it's very
    verbose!

---

    fix: log response correctly when not using json


mhkarimi1383:
Thanks for fixing logs :)

mhkarimi1383 (approved):

mhkarimi1383:
@mergify queue

PR #232, opened 2025-10-16
:sparkles: Add Volume source cloning

mhkarimi1383:
Also temporary snapshot support has been added for later usages

### DIFF
--- a/rawfile/rawfile_servicer.py
+++ b/rawfile/rawfile_servicer.py
@@ -1,10 +1,10 @@
 from pathlib import Path
 from subprocess import CalledProcessError
 
+import consts
 import grpc
 from config import config
 from internal_svc import SIGNATURE_METADATA
-from utils.errors import VolumeCloningNotSupported
 import utils.rawfile
 from consts import (
     FORMAT_OPTIONS_KEY,
@@ -16,7 +16,7 @@ from consts import (
 )
 from internal import internal_pb2, internal_pb2_grpc
 from csi import csi_pb2, csi_pb2_grpc
-from utils.rawfile import be_absent, be_symlink
+from utils.rawfile import be_absent, be_symlink, metadata
 from google.protobuf.wrappers_pb2 import BoolValue
 from orchestrator.k8s import node_ip_mapping, volume_to_node
 from utils.remote import get_capacity, init_rawfile, scrub
@@ -30,6 +30,7 @@ from utils.rawfile import (
 from utils.devices import mountpoint_to_dev, device_stats
 from utils.units import normalize_parameters, str_to_bool
 from analytics.ga4 import send_event, Usage
+from utils.snapshot_manager import manager as snapshot_manager
 
 NODE_NAME_TOPOLOGY_KEY = "hostname"
 
@@ -170,6 +171,7 @@ class RawFileControllerServicer(csi_pb2_grpc.ControllerServicer):
                 Cap(rpc=Cap.RPC(type=Cap.RPC.EXPAND_VOLUME)),
                 Cap(rpc=Cap.RPC(type=Cap.RPC.CREATE_DELETE_SNAPSHOT)),
                 Cap(rpc=Cap.RPC(type=Cap.RPC.LIST_SNAPSHOTS)),
+                Cap(rpc=Cap.RPC(type=Cap.RPC.CLONE_VOLUME)),
             ]
         )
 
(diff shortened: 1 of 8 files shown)


rawfile/tests/smoke.feature
```diff
@@ -26,17 +26,44 @@ Feature: Basic Functionality
...
+    Then source PVC should be eventually be deleted
     When we delete the snapshot
-    Then it should be eventually be deleted
+    Then snapshot should be eventually be deleted
+
+  Examples:
+    | copy_on_write | fsfreeze |
+    | false         | false    |
+    | true          | false    |
+    | false         | true     |
```
tiagolobocastro, line 49: true true seems to be missing from both tables

rawfile/utils/logs.py
```diff
@@ -113,27 +113,29 @@ def wrap(self, request, context):
...
             )
-            logger.info("GRPC Server Access Log", **args)
+            logger.success("GRPC Server Access Log", **args)
             return res
         except Exception as exc:
             end = datetime.now()
             args.update(
                 {
-                    "response": None,
+                    "response": (MessageToDict(res) if is_json else res)
```
tiagolobocastro, line 135: can we even has res here?

rawfile/rawfile_servicer.py
```diff
@@ -212,8 +222,25 @@ def CreateVolume(self, request: csi_pb2.CreateVolumeRequest, context):
...
+                source_meta = metadata(request.volume_content_source.volume.volume_id)
+                current_snapshot = snapshot_manager.list_snapshots(
+                    volume_id=request.volume_content_source.volume.volume_id,
+                    snapshot_name=f"{request.volume_content_source.volume.volume_id}-clone-source",
+                    temporary=True,
+                )
+                if current_snapshot.data and current_snapshot.data[0].ready:
+                    snapshot_id = current_snapshot.data[0].snapshot_id
+                else:
+                    snapshot_id = snapshot_manager.create_snapshot(
```
tiagolobocastro, line 235: so we create a snapshot during the call here?
and then we clone the snapshot 
and then we delete the snapshot?

Could you please extend the other document with this behavior as well?
Also I wonder, can not simply create the snapshot as the clone? I mean the snapshot *is* the clone right? Without `reflink` this would avoid copying twice?

mhkarimi1383: Yeah we create a temporary snapshot and then we just create a full clone out of that and we remove it

That's because we don't want to have a connection between destination volume and source volume to be able to remove them independently

I have to add another design doc for that if needed

rawfile/tests/smoke.feature
mhkarimi1383: Yeah I thought that this is not needed, but I will add it

rawfile/utils/logs.py
mhkarimi1383: If other parts have failed after calling the handler, it should be less than 1% I think but I added to be ensured that we cover that case

rawfile/rawfile_servicer.py
tiagolobocastro: > That's because we don't want to have a connection between destination volume and source volume to be able to remove them independently

What do you mean by a connection?
What I meant is, what we are copying into, we could just use that as the volume? Ex: `cp source-volume.img new-volume.img`

rawfile/tests/smoke.feature
tiagolobocastro: Yeah would be good for completeness, thanks

rawfile/rawfile_servicer.py
mhkarimi1383: > What do you mean by a connection?

Reflink or something like that

> What I meant is, what we are copying into, we could just use that as the volume? Ex: `cp source-volume.img new-volume.img`

That's because we should be able to clone in used volumes if possible
a 3 step process will protect us when doing such things, for example if we are allowed to reflink, since we want to do full independent full clone we can create a reflink snapshot and then clone that to the destination

tiagolobocastro: Right, in case of `reflink` yes that makes sense, my case was more for non-`cow`, where you have to deep copy twice.

mhkarimi1383: Yeah, But I was just like not making it conditional, It can get complex

tiagolobocastro: Yeah we'd need another entity to determine which is the best approach depending on the parameters.
We could consider doing it as another improvement.

tiagolobocastro (approved):

rawfile/rawfile_servicer.py
mhkarimi1383: Yeah, I will merge this for now

mhkarimi1383:
@mergify squash

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify queue

PR #235, opened 2025-10-20
:wrench: Add kubelet path configuration

### LINKED ISSUE #233: k0s Support
### Describe the problem/challenge you have

K0s uses a non standard path for it's kubelet

### Describe the solution you'd like

We should be able to customize `/var/lib/kubelet/`

### Anything else you would like to add?

```
MountVolume.SetUp failed for volume "registration-dir" : hostPath type check failed: /var/lib/kubelet/plugins_registry is not a directory

(⎈ | user@local:) λ  ls /var/lib/k0s/kubelet                                                                                                                                                                                                 ~
 checkpoints   pki   plugins   plugins_registry   pod-resources   pods   actuated_pods_state   allocated_pods_state   cpu_manager_state   dra_manager_state   memory_manager_state                                             /0.1s
```

mhkarimi1383:
Resolves #233

### DIFF
--- a/deploy/helm/rawfile-localpv/values.yaml
+++ b/deploy/helm/rawfile-localpv/values.yaml
@@ -129,6 +129,9 @@ node:
   # -- Data dir path for provisioner to be used by provisioner
   dataDirPath: /var/csi/rawfile
 
+  # -- Kubelet path (Set to `/var/lib/k0s/kubelet` for k0s)
+  kubeletPath: /var/lib/kubelet
+
   # -- Sets compute resources for node component
   resources:
     {}
--- a/deploy/helm/rawfile-localpv/README.md
+++ b/deploy/helm/rawfile-localpv/README.md
@@ -73,6 +73,7 @@ Please follow the [install guide](https://github.com/openebs/rawfile-localpv/tre
 | node.image.tag | string | `""` | Overrides default image tag for node component |
 | node.internalGRPC.port | int | `4500` | Port Number used for internal communication gRPC server |
 | node.internalGRPC.workers | int | `10` | gRPC worker count used for internal communication |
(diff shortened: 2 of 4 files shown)


deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -129,6 +129,9 @@ node:
   # -- Data dir path for provisioner to be used by provisioner
   dataDirPath: /var/csi/rawfile
 
+  # -- Kubelet path (Set to `/var/lib/k0s/kubelet` for k0s) without trailing slash
+  kubeletPath: /var/lib/kubelet
```
tiagolobocastro, line 133: I suggest we add a template function to generate a `kubeletPath` without the suffix and use that instead of the var directly

tiagolobocastro (approved):

deploy/helm/rawfile-localpv/values.yaml
mhkarimi1383: Can you give me an example from other local PVs, I mean we are depending on the real path of the kubelet in some places

tiagolobocastro: Example on the [zfs](https://github.com/openebs/zfs-localpv/blob/85be99d1d61e977d25f4fa52e58def80659b420f/deploy/helm/charts/templates/_helpers.tpl#L160).
Though not sure why it's keeping the slash at the end, guess depends on how it's used

mhkarimi1383: Oh got it

Done.

tiagolobocastro (approved):

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify queue

PR #238, opened 2025-10-22
:recycle: Use volume manager for volume lifecycle and move them to background

mhkarimi1383:


rawfile/utils/lock.py
```diff
@@ -7,15 +7,22 @@
...
 
     def __enter__(self):
-        # grabs the lock or fails right away
-        fcntl.flock(self.file, fcntl.LOCK_EX | fcntl.LOCK_NB)
-        return self
+        while True:
+            try:
+                fcntl.flock(self.file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                return self
+            except Exception as e:
```
tiagolobocastro: Should we loop only in case of loosing the lock, and bail out for other errors, ex: file not found

rawfile/utils/lock.py
```diff
@@ -7,15 +7,22 @@
 
 
 class VolLock:
-    def __init__(self, volume_id):
+    def __init__(self, volume_id, clear_on_releae: bool = True, wait: bool = False):
```
tiagolobocastro: typos on release here

rawfile/utils/lock.py
```diff
@@ -24,3 +31,5 @@ def release(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.release()
+        if self.clear_on_releae:
+            self.path.unlink(missing_ok=True)
```
tiagolobocastro, line 35: what's the use case for this?

rawfile/utils/volume_manager.py
```diff
@@ -0,0 +1,229 @@
...
+        source_type: VolumeSource | None = None,
+        source_id: str | None = None,
+    ):
+        if source_id and (not source_type):
+            raise SourceTypeRequired(source_id)
+        snapshot_name = None
+        source_volume_id = None
+        img_data_dir = img_dir(volume_id)
+        img_data_dir.mkdir(mode=consts.D_PERMS, exist_ok=True)
+        patch_metadata(volume_id, {"ready": False})
```
tiagolobocastro: Do we need this for parentless volumes?

rawfile/rawfile_servicer.py
```diff
@@ -298,18 +269,25 @@ def volume_provision(usage: Usage):
...
-        try:
-            size = scrub(volume_id=request.volume_id)
-        except CalledProcessError as exc:
-            if exc.returncode == VOLUME_IN_USE_EXIT_CODE:
-                context.abort(grpc.StatusCode.FAILED_PRECONDITION, "Volume in use")
-            else:
-                raise exc
+        img_dir = utils.rawfile.img_dir(request.volume_id)
+        if not img_dir.exists():
+            return 0
```
tiagolobocastro:
```suggestion
            return csi_pb2.DeleteVolumeResponse()
```

rawfile/rawfile_servicer.py
```diff
@@ -298,18 +269,25 @@ def volume_provision(usage: Usage):
...
+
+        img_file = utils.rawfile.img_file(request.volume_id)
```
tiagolobocastro: any reason to move this here? Seemed a bit tidier in a separate function

rawfile/rawfile_servicer.py
```diff
@@ -254,29 +205,49 @@ def CreateVolume(self, request: csi_pb2.CreateVolumeRequest, context):
...
+                size,
+                thin_provision,
+                freezefs,
+                copy_on_write,
+                source_type,
+                source_id,
+            )
+        start_time = time.time()
+        while time.time() - start_time < 30:
+            is_ready = metadata_or(request.name).get("ready", False)
```
tiagolobocastro, line 249: Do we need to check for deletion?

```suggestion
            metadata = metadata_or(request.name)
            is_ready = metadata.get("ready", False)
            deleted_at = metadata.get("deleted_at", None)
            if deleted_at:
                context.abort....
```

rawfile/rawfile_servicer.py
```diff
@@ -254,29 +205,49 @@ def CreateVolume(self, request: csi_pb2.CreateVolumeRequest, context):
...
+                source_id = request.volume_content_source.snapshot.snapshot_id
+            elif all(
+                (
+                    request.volume_content_source.volume,
+                    request.volume_content_source.volume.volume_id,
                 )
+            ):
+                source_type = VolumeSource.volume
+                source_id = request.volume_content_source.volume.volume_id
+        if utils.storage_pool.get_capacity() < size:
```
tiagolobocastro: I wonder if this needs to change now
Image the scenario where we are copying in the background, and the capacity in the system reduces (ie perhaps even caused by our copying) and when the csi retries it fails here even though we're now ready.

Also does this size need to take into account the extra snapshot copy for non-cow?

rawfile/utils/volume_manager.py
mhkarimi1383: Since the entire creation is in the background, I have added this for all of the volumes

rawfile/utils/lock.py
mhkarimi1383: two reasons:

1. Checking for locks manually by user or us when trying to tshoot
2. Extra unlink requires extra import that causes import recursion (When removing volume)

rawfile/rawfile_servicer.py
mhkarimi1383: this is not the source, why checking the deletion?

rawfile/rawfile_servicer.py
mhkarimi1383: I will make it to check for 3x space requirement for the volume source

rawfile/rawfile_servicer.py
mhkarimi1383: Yeah I will move that

rawfile/rawfile_servicer.py
mhkarimi1383: Done.

rawfile/rawfile_servicer.py
mhkarimi1383: Done.

rawfile/rawfile_servicer.py
tiagolobocastro: In case creating is taking too long (non-cow) and user decides to delete the pvc? Though I guess we'd also need to cancel the creation task itself

mhkarimi1383: In that case it will call delete volume and it will get locked

mhkarimi1383: I think our volume deletion process is buggy I will change it a bit, in terms of lock management and things like that

rawfile/utils/lock.py
mhkarimi1383: Fixed.

rawfile/utils/lock.py
mhkarimi1383: Fixed

rawfile/rawfile_servicer.py
mhkarimi1383: At first we are taking lock
then we remove data dir and data
and after that we remove and release the lock

I will change the order a bit and the delete mark should happen with the lock

mhkarimi1383: Done.

I have done some protections

mhkarimi1383:
@mergify rebase

rawfile/tests/test_smoke.py
```diff
@@ -608,8 +608,9 @@ def _(source_writer_pod, source_pvc):
         )
     else:
         try:
-            logger.info(f"Deleting snapshot writer POD: {pod_name}")
+            logger.info(f"Deleting writer POD: {pod_name}")
             client.CoreV1Api().delete_namespaced_pod(pod_name, namespace)
+            wait_pod_deleted(pod_name)
```
tiagolobocastro: it's below already on 617?

rawfile/rawfile_servicer.py
```diff
@@ -205,78 +190,70 @@ def CreateVolume(self, request: csi_pb2.CreateVolumeRequest, context):
...
-                )
+        start_time = time.time()
+        while time.time() - start_time < 30:
+            is_ready = metadata_or(request.name).get("ready", False)
+            if is_ready:
+                break
+            time.sleep(0.5)
+
+        if not is_ready:
+            context.abort(grpc.StatusCode.DEADLINE_EXCEEDED, "Volume Still Creating...")
```
tiagolobocastro, line 255: nit:
```suggestion
            context.abort(grpc.StatusCode.DEADLINE_EXCEEDED, "Volume still creating...")
```

tiagolobocastro (approved):

rawfile/tests/test_smoke.py
mhkarimi1383: I will revert this

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify squash

mhkarimi1383:
@mergify queue

PR #247, opened 2025-11-09
:recycle: Separating metadata dir and data dir

### LINKED ISSUE #234: Feature: Multiple StorageClasses with separate dataDirPaths
I've got a fast and a slow storage filesystem mounted on my node, and I'd like to offer storageclasses for each. Right now, it looks like rawfile-localpv only supports one global dataDirPath?

mhkarimi1383:
Partially fixes #234

### DIFF
--- a/deploy/helm/rawfile-localpv/values.yaml
+++ b/deploy/helm/rawfile-localpv/values.yaml
@@ -129,6 +129,9 @@ node:
   # -- Data dir path for provisioner to be used by provisioner
   dataDirPath: /var/csi/rawfile
 
+  # -- Metadata dir path for rawfile volumes metadata and tasks store file
+  metadataDirPath: /var/local/openebs/rawfile/meta
+
   # -- Kubelet path (Set to `/var/lib/k0s/kubelet` for k0s)
   kubeletPath: /var/lib/kubelet
 
--- a/rawfile/rawfile.py
+++ b/rawfile/rawfile.py
@@ -22,17 +22,32 @@ from utils.volume_manager import manager as volume_manager
 import os
 
 
-def node_driver_preflight_checks():
+def __create_and_check_directory(dir: Path):
+    if not dir.exists():
+        logger.info("Creating directory", path=str(dir))
+        dir.mkdir(parents=True, exist_ok=True)
+    if not dir.is_dir():
+        raise RuntimeError(f"{dir} is not a directory")
+    if not os.access(dir, os.W_OK | os.R_OK | os.X_OK):
+        raise RuntimeError(f"{dir} is not accessible")
+    dir.chmod(consts.D_PERMS)
+
+
+def node_driver_preflight_checks(task_manager: task_manager.TaskManager):
+    if not config.csi_driver:
+        raise RuntimeError("CSI Driver configuration is missing")
+    if not config.csi_driver.metadata_dir:
+        raise RuntimeError("Metadata directory is not set for node plugin")
     data_dir = Path(consts.DATA_DIR)
-    if not data_dir.exists():
-        logger.info("Creating data directory", path=str(data_dir))
-        data_dir.mkdir(parents=True, exist_ok=True)
-    if not data_dir.is_dir():
-        raise RuntimeError(f"{data_dir} is not a directory")
-    if not os.access(data_dir, os.W_OK | os.R_OK | os.X_OK):
-        raise RuntimeError(f"{data_dir} is not accessible")
-    data_dir.chmod(consts.D_PERMS)
+    dirs = (
+        config.csi_driver.metadata_dir,
+        data_dir,
+    )
+    for dir in dirs:
+        __create_and_check_directory(dir)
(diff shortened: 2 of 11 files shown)


mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify rebase

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -129,6 +129,9 @@ node:
   # -- Data dir path for provisioner to be used by provisioner
   dataDirPath: /var/csi/rawfile
 
+  # -- Metadata dir path for rawfile volumes
+  metadataDirPath: /var/csi/rawfile/metadata
```
tiagolobocastro: What happens on upgrades from current version?

mhkarimi1383: I have added a migration process

mhkarimi1383: I will run before schema migration

rawfile/rawfile.py
```diff
@@ -22,17 +22,31 @@
...
-        raise RuntimeError(f"{data_dir} is not accessible")
-    data_dir.chmod(consts.D_PERMS)
+    dirs = (
+        config.csi_driver.metadata_dir,
+        data_dir,
+    )
+    for dir in dirs:
+        __create_and_check_directory(dir)
     volume_manager.migrate_all_volume_schemas()
+    volume_manager.migrate_metadata_dir()
```
mhkarimi1383, line 48: This one should be moved before migration

tiagolobocastro: You mean migrating metadata before "all_volumes_schemas" ?

rawfile/utils/volume_manager.py
```diff
@@ -256,6 +257,20 @@ def migrate_metadata(self, volume_id, target_version):
...
+            for f in (
+                "disk.meta",
+                "disk.meta.tmp",
+                "disk.lock",
+            ):
+                src = Path(f"{consts.DATA_DIR}/{volume_id}/{f}")
+                if src.exists():
+                    dst = meta_dir(volume_id) / f
+                    dst.parent.mkdir(parents=True, exist_ok=True)
+                    src.rename(dst)
```
tiagolobocastro: this will only work if the filesystem is the same?

tiagolobocastro (commented):
Also how do you envision adding the storage pools to the configuration?

rawfile/rawfile.py
mhkarimi1383: Yeah
Should migrate the directory before the schema

rawfile/utils/volume_manager.py
mhkarimi1383: Yeah looks like we should use shutil.move

mhkarimi1383:
> Also how do you envision adding the storage pools to the configuration?
> 

a new config option called storagePools

It will be an array consist of

Name that should be unique
Path
isDefault (only one default pool is allowed)
capacityOverride
reservedCapacity

In the chart we will add the current data dir as a default pool

And a new storageclass parameter called storagePool

tiagolobocastro (approved):
btw typo on commit and PR -> s/Seperating/Separating

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify squash

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -129,6 +129,9 @@ node:
   # -- Data dir path for provisioner to be used by provisioner
   dataDirPath: /var/csi/rawfile
 
+  # -- Metadata dir path for rawfile volumes
```
tiagolobocastro: Does it make sense to start using /var/local which we use on the dynamic, ex: /var/local/openebs/rawfile ?
Though this would be a real breaking change

tiagolobocastro (approved):

deploy/helm/rawfile-localpv/values.yaml
mhkarimi1383: I want to change this too, but it will introduce a breaking change, we can just make this change with the storage pool, and use this address for default pool

mhkarimi1383: /var/local/openebs/rawfile/data
/var/local/openebs/rawfile/meta

mhkarimi1383: If that's Okay

I will set the default path of metadata for now here
And we change datadir path within the storage pool support

tiagolobocastro: Sure, sounds good

mhkarimi1383: Done.

mhkarimi1383:
@mergify squash

mhkarimi1383:
@mergify queue

PR #256, opened 2025-11-24
ci: push rawfile-localpv helm chart as oci

tiagolobocastro:


mhkarimi1383 (approved):

mhkarimi1383:
Should we also publish image in the GHCR too?

mhkarimi1383:
Can you also update docs to include it?

tiagolobocastro:
> Should we also publish image in the GHCR too?

Not sure, is there a benefit? Tbh the only reason why I used ghcr for the helm chart is because we did it on openebs as ghcr supports sub-namespaces whilst docker-hub doesn't.

So also here we should ask, should we publish OCI chart to [dockerhub](https://github.com/openebs/openebs/issues/3742#issuecomment-3571742317) as well?

mhkarimi1383:
> > Should we also publish image in the GHCR too?
> 
> Not sure, is there a benefit? Tbh the only reason why I used ghcr for the helm chart is because we did it on openebs as ghcr supports sub-namespaces whilst docker-hub doesn't.
> 
> So also here we should ask, should we publish OCI chart to [dockerhub](https://github.com/openebs/openebs/issues/3742#issuecomment-3571742317) as well?

 Oh, I see
For that we will need another org in dockerhub, called openebs-charts for example (I think Bitnami is doing the same)

tiagolobocastro:
> > > Should we also publish image in the GHCR too?
> > 
> > 
> > Not sure, is there a benefit? Tbh the only reason why I used ghcr for the helm chart is because we did it on openebs as ghcr supports sub-namespaces whilst docker-hub doesn't.
> > So also here we should ask, should we publish OCI chart to [dockerhub](https://github.com/openebs/openebs/issues/3742#issuecomment-3571742317) as well?
> 
> Oh, I see For that we will need another org in dockerhub, called openebs-charts for example (I think Bitnami is doing the same)

Going with this approach for now.
If need be, then we'll have to setup another org then

tiagolobocastro:
@mergify queue

PR #282, opened 2026-02-03
Fix reserving capacity logic

pasha-gurkov:
Summary: fix the reserving capacity logic and make it a little bit more flexible.

Deprecate capacity_override

Enhance reserved_capacity calculations with reserved_capacity_mode

Fix the thin provisioning capacity calculations

Add account_volumes_free_space flag

Modify charts accordingly

Rationale:
Currently, capacity calculations suffer from several problems:

1. capacity_override doesn't do anything; the only place override is used at is `get_capacity()` function, which only cares about device free space calculated directly from `os.statvfs()` call.

2. `get_capacity()` reserves capacity based on the free space left on the device, which is peculiar. It should be based on total capacity instead.

3. `get_capacity()` always assumes thin provisioning, and prevents from overprovisioning. Both the assumptions may be wrong.

New flag, `account_volumes_free_space` fixes point no. 3: if not set, the free space from `os.statvfs()` is the actual free space on disk, which works for thick provisioning and thin provisioning with overprovisioning. If set, it prohibits overprovisioning if volumes are thin-provisioned. This approach is not flawless, as it won't work properly if a single device or storage pool mixes different provisioning type volumes, but fixing that further is out of the scope of this commit.

`get_capacity()` and `path_stats()` changes address points 1 and 2.

`reserved_capacity_mode` allows user to choose whether they want to limit capacity available to CSI driver in top-down or bottom-up manner. For example, if the disk is 1TiB large and `reserved_capacity` is set to 100GiB and the mode is `plain` (default), CSI driver will have access to 1TiB - 100GiB = 900GiB. If the mode is `subtract_from_total`, this is reversed: CSI driver will have access to 1TiB - (1TiB - 100GiB) = 100GiB. This allows to cap CSI driver using absolute numbers while keeping the change backward compatible. Same goes for percentages.

I didn't find proper way to generate chart's README.md though so that's outdated. I would appreciate pointers there.

### DIFF
--- a/deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
+++ b/deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
@@ -81,10 +81,6 @@ spec:
             - name: RESERVED_CAPACITY
               value: {{ .Values.reservedCapacity | toString | quote }}
             {{- end }}
-            {{- if .Values.capacityOverride }}
-            - name: CAPACITY_OVERRIDE
-              value: {{ .Values.capacityOverride | toString | quote }}
-            {{- end }}
             - name: CSI_DRIVER__GRPC_WORKERS
               value: {{ .Values.node.grpcWorkers | toString | quote }}
             - name: GA_ENABLED
@@ -136,9 +132,6 @@ spec:
                     {{- with .Values.reservedCapacity }}
                     , "reserved_capacity": "{{ . | toString }}"
                     {{- end }}
-                    {{- with .Values.capacityOverride }}
-                    , "capacity_override": "{{ . | toString }}"
-                    {{- end }}
                     }
                   {{- else }}
                   {{- $keys := keys .Values.node.storagePools | sortAlpha }}
@@ -148,8 +141,13 @@ spec:
                     {{- with (index $.Values.node.storagePools $name).reservedCapacity }}
                     , "reserved_capacity": "{{ . | toString }}"
                     {{- end }}
-                    {{- with (index $.Values.node.storagePools $name).capacityOverride }}
-                    , "capacity_override": "{{ . | toString }}"
+                    {{- with (index $.Values.node.storagePools $name).reservedCapacityMode }}
+                    , "reserved_capacity_mode": "{{ . | toString }}"
+                    {{- end }}
+                    {{- with (index $.Values.node.storagePools $name) }}
+                    {{- if hasKey . "accountVolumesFreeSpace" }}
+                    , "account_volumes_free_space": {{ .accountVolumesFreeSpace }}
+                    {{- end }}
                     {{- end }}
                   }{{ if lt (add1 $i) (len $keys) }},{{ end }}
                   {{- end }}
--- a/deploy/helm/rawfile-localpv/values.yaml
+++ b/deploy/helm/rawfile-localpv/values.yaml
@@ -8,8 +8,6 @@ logFormat: json
 
 # -- Used to reserve capacity on each node for data dir storage on each host (Supports percentage and size) [e.g. `25%` or `50GB` or `10MiB`] (Deprecated, use storagePools.reservedCapacity instead)
 reservedCapacity: ""
-# -- Overrides total capacity of the storage for data dir storage on each host (Support size values) [e.g. `50GB` or `10MiB`] (Deprecated, use storagePools.capacityOverride instead)
-capacityOverride: ""
 
 auth:
   # -- Enables authentication for internal gRPC server
@@ -158,7 +156,7 @@ node:
   # -- Default filesystem type for rawfile volumes (Currently supports `btrfs`, `xfs` and `ext4` [which is default])
   defaultFs: ext4
 
-  # -- Path to store data dir (Depricated, use storagePools.path instead)
+  # -- Path to store data dir (Deprecated, use storagePools.path instead)
   dataDirPath: /var/csi/rawfile
 
   # -- Default storage pool name
@@ -167,8 +165,10 @@ node:
   # storagePools:
   #   default:
   #     path: /var/local/openebs/rawfile/default-pool/
+  # -- Reserved capacity outside of the storage pool, can be specified in bytes or percentage of the underlying device capacity
   #     reservedCapacity: ""
-  #     capacityOverride: ""
+  # -- Reserved capacity mode. `Plain` means it's taken as-is, `subtract-from-total` means it's subtracted from underlying device capacity. Example: if reservedCapacity is 100GiB, then `plain` means the pool can grow until 100GiB is left, and `subtract-from-total` means it can grow up to 100GiB. Default: `plain`
+  #     reservedCapacityMode: "plain" | "subtract-from-total"
 
   # -- Metadata dir path for rawfile volumes metadata and tasks store file
   metadataDirPath: /var/local/openebs/rawfile/{{ .Release.Name }}/meta
--- a/rawfile/config/model.py
+++ b/rawfile/config/model.py
@@ -1,5 +1,12 @@
 import re
 import warnings
+import consts
+import json
+
+from filesystem.types import FileSystemName
+from typing import Annotated, Final, Literal
+from datetime import timedelta
+
 from pydantic_settings import (
     BaseSettings,
     CliSettingsSource,
@@ -7,7 +14,6 @@ from pydantic_settings import (
     PydanticBaseSettingsSource,
     SettingsConfigDict,
 )
-from filesystem.types import FileSystemName
 from pydantic import (
     AliasChoices,
     AnyUrl,
@@ -20,11 +26,9 @@ from pydantic import (
     model_validator,
 )
 from pydantic.networks import IPvAnyAddress
-from typing import Annotated, Final, Literal
-import consts
+
 from utils.logs import LoggingFormats
-from datetime import timedelta
-import json
+from utils.modeltypes import ReservedCapacityMode
 
 NAME_REGEX: Final[re.Pattern] = re.compile(
     r"^(?=.{1,253}$)(?!.*\.\.)([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?)(\.([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?))*$"
@@ -51,18 +55,26 @@ class Capabilities(BaseModel):
 class StoragePool(BaseModel):
     path: DirectoryPath = Field(description="Path of the pool, Should be unique")
     reserved_capacity: (
-        ByteSize
-        | Annotated[
+        Annotated[
             str,
             StringConstraints(strip_whitespace=True, pattern=r"^\d+%$"),
         ]
+        | ByteSize
     ) = Field(
         default=ByteSize(0),
-        description="Reserves capacity of storage pool",
+        description="""Reserves some disk capacity for uses other than this driver. May be specified in percentage
+        of total capacity or in bytes with commonly used suffixes""",
+        # since the ByteSize parses '123%', union_mode='left_to_right' and
+        # annotated string coming first in the union matters
+        union_mode="left_to_right",
     )
-    capacity_override: ByteSize | None = Field(
-        default=None,
-        description="Overrides total capacity of storage pool",
+
+    reserved_capacity_mode: ReservedCapacityMode = Field(
+        default=ReservedCapacityMode.PLAIN,
+        description="""Defines reserved mode. `plain` means reserved capacity is set directly as-is;
+        `subtract-from-total` means effective reserved capacity is total capacity minus reserved_capacity.
+        Effectively, `subtract-from-total` is analogous to "reserve for this pool only", while "plain"
+        is "reserve for everything else but this pool".""",
     )
 
 
--- a/rawfile/metrics.py
+++ b/rawfile/metrics.py
@@ -20,6 +20,7 @@ class VolumeStatsCollector(object):
     def __init__(self, node):
         self.node = node
 
+    # TODO (pgu, 09.02.2026): should these metrics be per node per pool?
     def collect(self):
         remaining_capacity = GaugeMetricFamily(
             "rawfile_remaining_capacity",
@@ -42,7 +43,7 @@ class VolumeStatsCollector(object):
         remaining_capacity.add_metric([self.node], get_remaining_capacity())
         for volume_id, stats in volume_manager.get_all_volumes_stats().items():
             volume_used.add_metric([self.node, volume_id], stats["used"])
-            volume_total.add_metric([self.node, volume_id], stats["total"])
+            volume_total.add_metric([self.node, volume_id], stats["logical_size"])
         return [remaining_capacity, volume_used, volume_total]
 
 
--- a/rawfile/utils/modeltypes.py
+++ b/rawfile/utils/modeltypes.py
@@ -1,0 +1,6 @@
+from enum import StrEnum
+
+
+class ReservedCapacityMode(StrEnum):
+    PLAIN = "plain"
+    SUBTRACT_FROM_TOTAL = "subtract-from-total"
--- a/rawfile/utils/storage_pool.py
+++ b/rawfile/utils/storage_pool.py
@@ -1,18 +1,71 @@
-from utils.devices import path_stats
-from utils.volume_manager import manager as volume_manager
+from pydantic import ByteSize
+
 from config import config
+from utils.devices import statvfs
+from utils.volume_manager import manager as volume_manager
+from utils.modeltypes import ReservedCapacityMode
+
+from collections.abc import Callable
+
+
+# if the mode is "plain", reserved capacity is taken as specified in config
+def plain_reserved_capacity_handler(
+    disk_total_size: int, reserved_capacity: ByteSize | str
+) -> int:
+    if isinstance(reserved_capacity, str):
+        return disk_total_size * int(reserved_capacity[:-1]) / 100
+    elif isinstance(reserved_capacity, ByteSize):
+        return reserved_capacity.to("B")
+
+
+# if the mode is "subtract-from-total", reserved capacity is taken as (total - specified in config)
+def subtract_from_total_reserved_capacity_handler(
+    disk_total_size: int, reserved_capacity: ByteSize | str
+) -> int:
+    if isinstance(reserved_capacity, str):
+        return disk_total_size - disk_total_size * int(reserved_capacity[:-1]) / 100
+    elif isinstance(reserved_capacity, ByteSize):
+        return disk_total_size - reserved_capacity.to("B")
+
+
+reserved_capacity_handlers: dict[
+    ReservedCapacityMode, Callable[[int, ByteSize | str], int]
+] = {
+    ReservedCapacityMode.PLAIN: plain_reserved_capacity_handler,
+    ReservedCapacityMode.SUBTRACT_FROM_TOTAL: subtract_from_total_reserved_capacity_handler,
+}
 
 
 def get_capacity(storage_pool: str | None = None):
     if not storage_pool:
         storage_pool = config.csi_driver.default_pool
     pool = config.csi_driver.storage_pools[storage_pool]
-    disk_free_size = path_stats(pool.path, pool.capacity_override)["fs_avail"]
-    capacity = disk_free_size
-    for volume_stat in volume_manager.get_all_volumes_stats().values():
-        capacity -= volume_stat["total"] - volume_stat["used"]
-    if isinstance(pool.reserved_capacity, str):
-        capacity -= capacity * int(pool.reserved_capacity[:-1]) / 100
-    else:
-        capacity -= pool.reserved_capacity.to("B")
-    return max(capacity, 0)
+    stats = statvfs(pool.path)
+    disk_total_size = stats["fs_size"]
+    disk_free_size = stats["fs_avail"]
+    reserved_capacity_config = pool.reserved_capacity
+    reserved_capacity_mode = pool.reserved_capacity_mode
+
+    # reserved_capacity is the capacity reserved for everything else but this pool
+    # so the capacity reserved for this pool is (disk_total_size - reserved_capacity)
+    reserved_capacity = reserved_capacity_handlers[reserved_capacity_mode](
+        disk_total_size, reserved_capacity_config
+    )
+
+    volumes_physical_size = 0
+
+    volume_stats_values = volume_manager.get_volumes_stats_by_pool(
+        storage_pool
+    ).values()
+    for volume_stat in volume_stats_values:
+        # if the backing file is sparse, count only the defacto allocated blocks
+        # if not, count the whole thing
+        volumes_physical_size += volume_stat["physical_size"]
+
+    capacity = max(
+        min(
+            disk_total_size - reserved_capacity - volumes_physical_size, disk_free_size
+        ),
+        0,
+    )
+    return capacity
--- a/rawfile/utils/volume_manager.py
+++ b/rawfile/utils/volume_manager.py
@@ -2,7 +2,7 @@ from enum import Enum
 import time
 from typing import TypedDict
 import consts
-from utils.devices import device_to_mountpoint, path_stats
+from utils.devices import device_to_mountpoint, statvfs, stat
 from utils.errors import SourceTypeRequired, VolumeInUseError, VolumeSourceIsNotReady
 from utils.lock import VolLock
 from config import config
@@ -34,8 +34,12 @@ class VolumeSource(int, Enum):
 
 
 class VolumeStats(TypedDict):
-    total: int
+    physical_size: int
+    logical_size: int
     used: int
+    pool: str
+    thin_provision: bool
+    thin: bool
 
 
 class VolumeManager:
@@ -119,6 +123,7 @@ class VolumeManager:
                         "thin_provision": thin_provision,
                         "freezefs": freezefs,
                         "copy_on_write": copy_on_write,
+                        "storage_pool": storage_pool,
                     },
                 )
                 img_file.touch()
@@ -251,6 +256,9 @@ class VolumeManager:
 
     def get_volume_stats(self, volume_id) -> VolumeStats | None:
         try:
+            meta = metadata_or(volume_id)
+            pool = meta.get("storage_pool", config.csi_driver.default_pool)
+            thin_provision = meta.get("thin_provision", False)
             file = img_file(volume_id=volume_id)
             loop_devs = attached_loops(file.as_posix())
             if not (loop_devs and len(loop_devs)):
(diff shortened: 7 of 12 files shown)


rawfile/config/model.py
```diff
@@ -34,18 +38,38 @@
...
+    reserved_capacity_mode: ReservedCapacityModeEnum = Field(
+        default=ReservedCapacityModeEnum.plain,
+        description="""Defines reserved mode. `plain` means reserved capacity is set directly as-is;
+        `subtract_from_total` means effective reserved capacity is total capacity minus reserved_capacity.
+        Effectively, `subtract_from_total` is analogous to "reserve only for this pool", while "plain"
+        is "reserve for everything else but this pool".""",
     )
+
+    account_volumes_free_space: bool = Field(
+        default=False,
```
mhkarimi1383: This one should be True by default

rawfile/utils/modeltypes.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class ReservedCapacityModeEnum(str, Enum):
```
mhkarimi1383: Make this to follow the pattern that we have for other Enums please,

Use StrEnum, and make values to be kebab-case please, it's easier to set in configs

and make keys to be All upper please

deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
```diff
@@ -137,9 +137,15 @@ spec:
...
                     , "reserved_capacity": "{{ . | toString }}"
                     {{- end }}
+                    {{- with (index $.Values.node.storagePools $name).reservedCapacityMode }}
+                    , "reserved_capacity_mode": "{{ . | toString }}"
+                    {{- end }}
                     {{- with (index $.Values.node.storagePools $name).capacityOverride }}
                     , "capacity_override": "{{ . | toString }}"
                     {{- end }}
+                    {{- with (index $.Values.node.storagePools $name).accountVolumesFreeSpace }}
+                    , "account_volumes_free_space": "{{ . | toString }}"
```
mhkarimi1383: This should be bare true or false

rawfile/utils/storage_pool.py
```diff
@@ -1,18 +1,50 @@
...
     pool = config.csi_driver.storage_pools[storage_pool]
-    disk_free_size = path_stats(pool.path, pool.capacity_override)["fs_avail"]
-    capacity = disk_free_size
-    for volume_stat in volume_manager.get_all_volumes_stats().values():
-        capacity -= volume_stat["total"] - volume_stat["used"]
-    if isinstance(pool.reserved_capacity, str):
-        capacity -= capacity * int(pool.reserved_capacity[:-1]) / 100
-    else:
-        capacity -= pool.reserved_capacity.to("B")
+    stats = path_stats(pool.path)
```
mhkarimi1383: Here, Can you make to functions and map enum value into the functions and just call them?

it will make it cleaner and it will separate functionalities 


like


```python
reserved_capacity_handlers: dict[ReservedCapacityModeEnum, callable[ByteSize | str]] = {
    ReservedCapacityModeEnum.plain: plain_reserved_capacity_handler
    ReservedCapacityModeEnum.subtract_from_total: subtract_from_total_reserved_capacity_handler
}
```

and use it with

```python
reserved_capacity_handlers[reserved_capacity_mode](reserved_capacity_config)
```

It will make it more cleaner and easier to read

thanks

rawfile/utils/modeltypes.py
mhkarimi1383: Also you can Remove Enum suffix from end

rawfile/config/model.py
pasha-gurkov: Could you please explain why? 

If it's True, the assumption is the user only uses thin provisioning and doesn't want to overprovision. If the volumes are thick-provisioned, setting this to True leads to very strange results that don't really make sense.

I feel that having the available capacity equal to `disk free space - reserved space` by default follows the principle of the least surprise.

tiagolobocastro: This should not be needed, we should take correct decision per volume (some may be thin, some thick) 
I think the problem is that we're using the wrong stats for the calculation, we're probably using the filesystem stats of the volume rather than the real allocation on the host:

```
❯ fallocate -l 1GiB /tmp/thick.io
❯ truncate -s 1GiB /tmp/thin.io
❯ du -h /tmp/*.io
1.1G	/tmp/thick.io
0	/tmp/thin.io

❯ dd if=/dev/urandom of=/tmp/thin.io bs=1M count=10
10+0 records in
10+0 records out
10485760 bytes (10 MB, 10 MiB) copied, 0.0248047 s, 423 MB/s

❯ du -h /tmp/*.io
1.1G	/tmp/thick.io
10M	/tmp/thin.io
```

rawfile/utils/modeltypes.py
pasha-gurkov: Done.

deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
pasha-gurkov: Fixed.

rawfile/utils/storage_pool.py
pasha-gurkov: Done.

rawfile/config/model.py
pasha-gurkov: @tiagolobocastro 
The correctness of decision here depends on whether the user wants to overprovision, or not.

If you'd rather give the user no choice and ban overprovisioning, or force it on the contrary, it's fine. There's no need for the switch then.

If overprovisioning is always on, then the implementation is the simplest: just always use what `os.statvfs()` returns. It won't account for allocated but unused blocks.

If overprovisioning is always off, it's a different story. We'd need a hacky way to determine if the file is sparse or not, because even a thick-provisioned file becomes sparse if one formats without `-E nodiscard`, so one can't rely on what's in `storageClass` (or perform convoluted parsing of the format options). So it'd be an `os.stat()` call for every backing file, comparing `.st_size` to `.st_blocks * 512`; if the former is larger, then it's a sparse file; and the calculations would differ on file per file basis.
However, this is something I hoped to avoid at least in this MR not to inflate the scope.
If I'm missing something, please let me know.

We can leave the switch in _and_ accomodate automatically for thin- and thick-provisioned files with a hack illustrated above; or leave the switch and impose implicit "all files for a given underlying device must have the same 'thickness'"; or get rid of the switch, and ban or force overprovisioning. What route would you like to take?

mhkarimi1383: We have added capacity override for that, so users were able to override the total capacity of the data dir (per pool)

I think we should consider that the user will not over provision by default, since it's a required thing in a Production environment, but users should be able to enable over provisioning per pool

tiagolobocastro: > If overprovisioning is always off, it's a different story. We'd need a hacky way to determine if the file is sparse or not, because even a thick-provisioned file becomes sparse if one formats without -E nodiscard, so one can't rely on what's in storageClass (or perform convoluted parsing of the format options). 

Good point, indeed discard would deallocate the rawfile.
I guess we could enforce that thick volumes should be mounted with no discard?

> So it'd be an os.stat() call for every backing file, comparing .st_size to .st_blocks * 512; if the former is larger, then it's a sparse file; and the calculations would differ on file per file basis.

So you mean if the volume is meant to be thick provisioned, then we could check the current size of the rawfile, and if it's lower than the expected size, we could reduce the capacity by the difference?

Though it seems that thick provisioning is not really working as expected because of the discard - it's not really thick at all!
If we really want thick volumes, then I think we ought to ensure they're not mounted with discard

What do you think?

pasha-gurkov: > We have added capacity override for that, so users were able to override the total capacity of the data dir (per pool)

Yeah, but capacity override didn't actually influence anything, as pointed out in https://kubernetes.slack.com/archives/CUAKPFU78/p1769698572407449

With the new `reserved_capacity_mode` switch, you're able to cap the pool at X GiB if it's set to `subtract-from-total`.

> I think we should consider that the user will not over provision by default, since it's a required thing in a 
Production environment, but users should be able to enable over provisioning per pool

The more I think about it, the less I understand what "overprovisioning per pool" should mean. Consider this scenario:
- I have two pools A and B, both 100GiB, A with overprovisioning and B without, backed by a single 200GiB device
- the logic should be, it seems, that I could create volumes on A exceeding pool's capacity, and I couldn't do that for B
- I create 2 100GiB thin volumes on A, and one on B
- I fill A volumes with data completely
- I can't write anything to B's volume anymore. I overprovisioned B with no intentions to do so.
- the logic from above formally holds, but with surprising results.

What I'm trying to say is, the moment I employ thin volumes/sparse files, I overprovision everything backed by the given device. The concept of enabling/disabling overprovisioning with calculation tweaks only works if
- every storage pool is backed by a different physical device
- there's nothing besides that pool accessing the device.

Even if we can somehow force the first assumption (which is not feasible in my mind), we cannot do anything about the second one.

Also, please take a look at the last paragraphs in this comment.

> I guess we could enforce that thick volumes should be mounted with no discard?

It is a format option rather than a mount option; and I personally think it's a rabbit hole of parsing text format options.

> So you mean if the volume is meant to be thick provisioned, then we could check the current size of the rawfile, and if it's lower than the expected size, we could reduce the capacity by the difference?

I rather mean we don't distinguish between thick and thin files, just check the `os.stat()` output, and adjust the calculations accordingly.
So if overprovisioning is not desired, decrease available capacity by `stat.st_size - stat.st_blocks * 512` for every known volume in the pool. `stat.st_size` would be equal to `stat.st_blocks * 512` for any thick volume or fully utilised thin volume; but would be larger than, for non-fully utilised thin volumes.

> Though it seems that thick provisioning is not really working as expected because of the discard - it's not really thick at all!

Right now yeah, first you have to specify `-E nodiscard` in the format options for the file to actually remain thick, but even then current calculations in the form of
```
volume_stat["total"] - volume_stat["used"]
```
_implicitly assume_ you're only working with thin provisioned files. For thick ones, it leads to bogus results.

BTW, another thing that crossed my mind while writing this comment:

I don't really think we should be accounting for thin files at all in any special way. Sparse files enable overprovisioning (or rather overcommitting) _by design_; currently the calculations are trying to fight the concept by accounting for blocks that are not yet written, but _could be_. The problem is, we're only able to do that in the scope we know about -- i.e. files we created. What happens on the rest of the filesystem is out of the scope, unknown and cannot be influenced in any way.

Suppose there's a totally empty 100GiB device, the driver creates a 100GiB thin backing file with nothing in it. Then something else on the node writes 50GiB to the disk (as the device blocks are actually free). We already overcommitted at this point, the user won't be able to write full 100GiB until something else frees the space on the device.

So **my actual proposition** is, we shouldn't compensate for the yet-unused-space for thin files at all. The behaviour should be as straightforward as possible: just use the FS stats for the calculation. If the user wants overprovisioning, they use thin files. If not, they use thick files, and that's it.

What do you think?

EDIT: just to add, not only that follows the principle of least surprise, but also streamlines the calculations logic, and gets rid of `account_volumes_free_space` boolean switch.

tiagolobocastro: > What I'm trying to say is, the moment I employ thin volumes/sparse files, I overprovision everything backed by the given device. 

Yeah I agree with this, we are always over provisioned unless using the thick files with the no discard hack

> So **my actual proposition** is, we shouldn't compensate for the yet-unused-space for thin files at all. The behaviour should be as straightforward as possible: just use the FS stats for the calculation. If the user wants overprovisioning, they use thin files. If not, they use thick files, and that's it.
> 
> What do you think?

When you say "FS stats", what do you mean here, the storage pool free space?
But we'd need to still account for our volumes, otherwise this would give you an over-commitment of infinity?

pasha-gurkov: Just to let you know, I was wrong with my calculations, they should be fixed, I'll come back with an update next week.

pasha-gurkov: So the changes in the latest commit are as follows:
- don't treat thinly provisioned volumes as special, always account for actually allocated blocks in the capacity calculations
- drop `account_volumes_free_space` flag, as it's not needed anymore
- fix the capacity arithmetics, they were just wrong, sorry
- ban multiple storage pools backed by the same filesystem.

I've added another function to invoke `stat()` call to get actual allocated blocks for the volume, and mixed that information in to VolumeStats along with the pool name (to filter by pool in capacity calculations as we're not really interested what other pools are holding), and the intended (specified in the config) and actual (as it is on disk) thin-provision flag. The latter two are not really used right now, but might be useful in the future.

pasha-gurkov:
I just found out that fsid thingy I wrote worked by accident, hold on reviewing that bit please.

pasha-gurkov:
Changed `f_fsid` to `st_dev`, should be more portable and universal now.

pasha-gurkov:
Rebased on top of develop.

cometship:
Does this PR also fix the problem that if you have 2 storage pools:
- on a 512GB disk
- on a 2TB disk 

that it always reports the available capacity for both as 512GB (the size of the default pool)?

mhkarimi1383 (approved):

mhkarimi1383:
Sorry for being late, I have reviewed the logic looks like it's ok

@tiagolobocastro, Can you also review?

rawfile/config/model.py
```diff
@@ -34,18 +38,31 @@
...
+        description="""Defines reserved mode. `plain` means reserved capacity is set directly as-is;
+        `subtract-from-total` means effective reserved capacity is total capacity minus reserved_capacity.
+        Effectively, `subtract-from-total` is analogous to "reserve for this pool only", while "plain"
+        is "reserve for everything else but this pool".""",
     )
+
     capacity_override: ByteSize | None = Field(
         default=None,
-        description="Overrides total capacity of storage pool",
+        description="Deprecated, unused, left for backwards compatibility",
```
tiagolobocastro: Shouldn't we also remove it from the values.yaml?

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -156,6 +156,7 @@ node:
   #   default:
   #     path: /var/local/openebs/rawfile/default-pool/
   #     reservedCapacity: ""
+  #     reservedCapacityMode: "plain" | "subtract-from-total"
```
tiagolobocastro, line 171: We should "document" here how the plain and subtract work and which one is the default?

rawfile/utils/volume_manager.py
```diff
@@ -251,18 +256,31 @@ def gc_all_volumes(self, dry_run=True):
...
+
+            # note that "thin" here might differ from metadata's
+            # "thin_provision": if the volume was created as thick but formatted
+            # with discarding, it's factually thin i.e. sparse
+            result = {
+                "physical_size": file_stats["physical_size"],
+                "logical_size": file_stats["logical_size"],
+                "used": filesystem_stats["fs_usage"],
+                "pool": pool,
+                "thin": (file_stats["physical_size"] < file_stats["logical_size"]),
```
tiagolobocastro: what does thin represent here? Ex: if a sparse file is fully allocated, it'd be represented as thin even though it's not?

I guess we could do this?

```suggestion
                "thin": (file_stats["physical_size"] < file_stats["logical_size"]) or thin_provision,
```

rawfile/metrics.py
```diff
@@ -20,6 +20,7 @@ class VolumeStatsCollector(object):
     def __init__(self, node):
         self.node = node
 
+    # TODO (pgu, 09.02.2026): should these metrics be per node per pool?
```
tiagolobocastro, line 23: Yeah, I suspect the csi `GetCapacity` is broken with this as is

tiagolobocastro (approved):
We should also add the nodiscard by default to thick volumes?

rawfile/metrics.py
mhkarimi1383: I think it should return sum of all pools

deploy/helm/rawfile-localpv/values.yaml
mhkarimi1383: I think so

rawfile/metrics.py
tiagolobocastro: But a volume cannot span multiple pools?

Let's say we have 2 pools, with 100GiB each, and user wants to create a 1500GiB volume
If `GetCapacity` returns 200GiB then it looks like you can create a 150GiB volume, whereas you can only create a 100GiB volume.

Also IIRC `GetCapacity` does pass the StorageClass parameters, so I think we should be able to return the capacity for the specified parameters.

Maybe we should have both total capacity (for prometheus, not for CSI), and then capacity per pool for the CSI?

pasha-gurkov:
> We should also add the nodiscard by default to thick volumes?

I don't know really, it feels like too much intelligence and edge cases to me. Like for ext4 it's `-E nodiscard`, for btrfs/xfs it's `-K`. How should the default work, be appended to the end of format options, or be completely overwritten if a user specifies their own format options? If it's the former, it's hidden logic the user can't get rid of even if they want to; if it's the latter, we should mention it in the docs anyway to make it obvious. I kinda feel it's better to not employ non-obvious logic.

rawfile/config/model.py
pasha-gurkov: Would that cause any backwards compatibility issues?

deploy/helm/rawfile-localpv/values.yaml
pasha-gurkov: Will do in a bit.

rawfile/metrics.py
pasha-gurkov: Wait, I'm confused. This comment is on prometheus metrics, not the CSI GetCapacity calls.
Summing all pools for prometheus metrics, in my opinion, would indeed lead to confusion, just how it is right now.

But I don't understand these bits:
> Also IIRC GetCapacity does pass the StorageClass parameters, so I think we should be able to return the capacity for the specified parameters.

I believe that's what we do, CSIStorageCapacity returns capacity per storage class (which is really per storage pool that class belongs to). If you have 2 storage classes per 1 pool, the same capacity will be reported twice, but I don't think that's something we can do anything about since it's a part of CSI mech, am I wrong there?

> Maybe we should have both total capacity (for prometheus, not for CSI), and then capacity per pool for the CSI?

Again I believe that's what it is right now; I'm arguing it shouldn't be total for prometheus metrics, since then metrics would lead to confusion just like @tiagolobocastro stated above.

rawfile/utils/volume_manager.py
pasha-gurkov: Maybe that's a bad naming there as a source of confusion. Basically, `thin_provision` here is how the file was provisioned originally, and `thin` is whether the file actually is sparse or not. We could rename it to `sparse`? Or `thin_actual`, IDK.

I'm not opposed to your edit though.

rawfile/config/model.py
tiagolobocastro: Since we're making it unused now I don't think so, as long as it's not referenced by anything else it shouldn't?
Or at the very least update/remove the helm docs

rawfile/utils/volume_manager.py
tiagolobocastro: Yeah, I guess we could name it `sparse` as well, though `thin_actual` also seems fine, whichever you prefer.

rawfile/metrics.py
tiagolobocastro: > Wait, I'm confused. This comment is on prometheus metrics, not the CSI GetCapacity calls.

ah my bad, I thought the CSI `GetCapacity` was calling this directly, but seems not

> I believe that's what we do, CSIStorageCapacity returns capacity per storage class (which is really per storage pool that class belongs to). If you have 2 storage classes per 1 pool, the same capacity will be reported twice, but I don't think that's something we can do anything about since it's a part of CSI mech, am I wrong there?

Yeah that will happen and I think that's fine since 2 storage classes are configured for the same pool.
I was thinking in the case where you have 2 pools, and 1 storage class referencing only one pool (ofc), and in this case we should return the capacity for the correct pool.
If that's already happening then by bad, I must have misread the code.

tiagolobocastro:
> > We should also add the nodiscard by default to thick volumes?
> 
> I don't know really, it feels like too much intelligence and edge cases to me. Like for ext4 it's `-E nodiscard`, for btrfs/xfs it's `-K`. How should the default work, be appended to the end of format options, or be completely overwritten if a user specifies their own format options? If it's the former, it's hidden logic the user can't get rid of even if they want to; if it's the latter, we should mention it in the docs anyway to make it obvious. I kinda feel it's better to not employ non-obvious logic.

I think that should be fine, we already have some per-filesystem logic.
I'd be in favour of always using `no discard` (or whatever it is for each filesystem) for thick volumes and we document this is the behaviour that actually makes it thick, and user should not try to override this, otherwise it defeats the whole point of having a thick volume.

pasha-gurkov:
> > > We should also add the nodiscard by default to thick volumes?
> > 
> > 
> > I don't know really, it feels like too much intelligence and edge cases to me. Like for ext4 it's `-E nodiscard`, for btrfs/xfs it's `-K`. How should the default work, be appended to the end of format options, or be completely overwritten if a user specifies their own format options? If it's the former, it's hidden logic the user can't get rid of even if they want to; if it's the latter, we should mention it in the docs anyway to make it obvious. I kinda feel it's better to not employ non-obvious logic.
> 
> I think that should be fine, we already have some per-filesystem logic. I'd be in favour of always using `no discard` (or whatever it is for each filesystem) for thick volumes and we document this is the behaviour that actually makes it thick, and user should not try to override this, otherwise it defeats the whole point of having a thick volume.

Yeah, but what if they override part of it? I mean, I would also use something like ext4's `-m 0` for every volume, since there's no need really to reserve anything for superuser on such filesystems. Would I then have to specify `["-E", "nodiscard", "-m", "0"]`, or just `["-m", "0"]` in format options? What if the I run an ancient version of mkfs.ext4 that doesn't support nodiscard flag? How would I drop the default if I do want `fallocate`, but don't want `-E nodiscard`?

Anything I can think of feels fragile and way too intelligent. I have a proposal: let's make an issue out of this and return to it later, and not extend the scope of this story.

rawfile/metrics.py
pasha-gurkov: Yeah, the only thing I mean here is, let's not sum all the pools up when reporting a prometheus metric. I do however believe that deserves its own issue/MR.

tiagolobocastro:
> Yeah, but what if they override part of it? I mean, I would also use something like ext4's `-m 0` for every volume, since there's no need really to reserve anything for superuser on such filesystems. Would I then have to specify `["-E", "nodiscard", "-m", "0"]`, or just `["-m", "0"]` in format options? 

Only `["-m", "0"]`, we'd automatically add the `nodiscard`.

> What if the I run an ancient version of mkfs.ext4 that doesn't support nodiscard flag?

Then thick is definitely not supported there.

> How would I drop the default if I do want `fallocate`, but don't want `-E nodiscard`?

Then perhaps we'd need another option?
Another option would be to simply expose a way to select fallocate or truncate, and removing thin_provisioning configuration altogether, letting the user select the mkfs options themselves.

> Anything I can think of feels fragile and way too intelligent. I have a proposal: let's make an issue out of this and return to it later, and not extend the scope of this story.

Yeah and I don't disagree with that, though it seems the only way of support thick volumes out of the box; the alternative is we simply do away of the thin altogether, and leave it up to the user to use "correct" fs options
Sure, tracking this in a separate issue that sounds good.

rawfile/config/model.py
pasha-gurkov: I just nuked it from everywhere. Too drastic?

deploy/helm/rawfile-localpv/values.yaml
pasha-gurkov: Added a couple of lines, does it look OK?

rawfile/utils/volume_manager.py
pasha-gurkov: Gotcha, renamed to "sparse"!

pasha-gurkov:
> tracking this in a separate issue that sounds good.

Got it, I'll create one for this, and another for prometheus metrics then, thank you. Let's continue discussions there!

Concerning this MR, @tiagolobocastro, @mhkarimi1383 do you have any more feedback, or we're good to squash & merge?

pasha-gurkov:
Rebased & squashed.

mhkarimi1383 (approved):

mhkarimi1383:
I'm Okay with changes

mhkarimi1383:
What about capacity override? It was a feature request :) 

CC: @tiagolobocastro

tiagolobocastro:
> What about capacity override? It was a feature request :)
> 
> CC: @tiagolobocastro

oh I wasn't remembering that, we're actually removing that feature altogether here... is someone actually using it?
Maybe we should send "feature usage" metrics from the ga analytics :)

mhkarimi1383:
Let me find the feature request for that

mhkarimi1383:
https://github.com/openebs/rawfile-localpv/issues/24

pasha-gurkov:
> What about capacity override? It was a feature request :)
> 
> CC: @tiagolobocastro

Erm, do we need that, since I removed the logic compensating for non-allocated blocks if thin-provisioning? The logic is now, if you want overprovisioning, use thin-provisioning, if you don't, use thick.

pasha-gurkov:
Rebased on top of develop.

rawfile/metrics.py
pasha-gurkov: Resolving this as opened a separate issue.

pasha-gurkov:
> Does this PR also fix the problem that if you have 2 storage pools:
> 
> * on a 512GB disk
> * on a 2TB disk
> 
> that it always reports the available capacity for both as 512GB (the size of the default pool)?

@cometship I wasn't aware this is a problem, but it might fix it. If you could test it and tell if it works, that'd be great!

tiagolobocastro:
> > What about capacity override? It was a feature request :)
> > CC: @tiagolobocastro
> 
> Erm, do we need that, since I removed the logic compensating for non-allocated blocks if thin-provisioning? The logic is now, if you want overprovisioning, use thin-provisioning, if you don't, use thick.

I think we can remove it - looking back at the user request what they wanted was to make creation size check use only `fs_avail` from the data dir path, which is not exactly the same as capacity override that we had?

mhkarimi1383:
> exactly

If that's true we can merge this PR, If nothing left can you also approve it

tiagolobocastro:
@mergify queue

cometship:
> > Does this PR also fix the problem that if you have 2 storage pools:
> > 
> > * on a 512GB disk
> > * on a 2TB disk
> > 
> > that it always reports the available capacity for both as 512GB (the size of the default pool)?
> 
> @cometship I wasn't aware this is a problem, but it might fix it. If you could test it and tell if it works, that'd be great!

Unfortunately it did not fix the problem :/ 

`CSIStorageCapacity` still reports same capacity for both.

I created a new issue: https://github.com/openebs/rawfile-localpv/issues/302

tiagolobocastro:
I think its because the storage-pool is not being selected, I've raised #303 to address this

PR #284, opened 2026-02-04
feat: Optionally disable individual containers

pasha-gurkov:
Add `enabled` flag to snapshot-related and resizer containers.

Rationale: if there's no user intention to create VolumeSnapshots or resize volumes, there's no reason to run additional containers.

### DIFF
--- a/deploy/helm/rawfile-localpv/values.yaml
+++ b/deploy/helm/rawfile-localpv/values.yaml
@@ -262,3 +262,11 @@ crds:
     volumeSnapshots:
       # -- Install Volume Snapshot CRDs
       enabled: true
+
+capabilities:
+  snapshots:
+    # -- Sets whether taking volume snapshots is enabled. Required for volume cloning. Runs externalSnapshotter and snapshotController containers.
+    enabled: true
+  resize:
+    # -- Sets whether volume resizing is enabled. If disabled, don't deploy controller statefulset
+    enabled: true
--- a/deploy/helm/rawfile-localpv/README.md
+++ b/deploy/helm/rawfile-localpv/README.md
@@ -29,6 +29,8 @@ Please follow the [install guide](https://github.com/openebs/rawfile-localpv/tre
 | auth.enabled | bool | `true` | Enables authentication for internal gRPC server |
 | auth.secretName | string | `""` | If managing secrets outside the chart, use this to reference the secret name; otherwise, leave empty. |
 | auth.token | string | `""` | Sets authentication token for internal gRPC server, will generate one if nothing provided |
+| capabilities.resize.enabled | bool | `true` | Sets whether volume resizing is enabled. If disabled, don't deploy controller statefulset |
+| capabilities.snapshots.enabled | bool | `true` | Sets whether taking volume snapshots is enabled. Required for volume cloning. Runs externalSnapshotter and snapshotController containers. |
 | capacityOverride | string | `""` | Overrides total capacity of the storage for data dir storage on each host (Support size values) [e.g. `50GB` or `10MiB`] (Deprecated, use storagePools.capacityOverride instead) |
 | controller.affinity | string | `nil` | Affinities for controller component |
 | controller.externalResizer.image.registry | string | `""` | Image registry for `csi-resizer` |
--- a/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
+++ b/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.capabilities.resize.enabled }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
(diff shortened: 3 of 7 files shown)


deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -116,6 +118,8 @@ node:
       repository: sig-storage/csi-snapshotter
       # -- Image Tag for `csi-snapshotter`
       tag: v8.2.1
+    # -- Whether to run externalSnapshotter
+    enabled: true
```
tiagolobocastro: Is there any use-case for having `externalSnapshotter` and `snapshotController` individually enabled or would a single enable be sufficient here?

tiagolobocastro (approved):

mhkarimi1383:
@pasha-gurkov 

That's great to be able to control features here, but first of all

@tiagolobocastro do we have this functionality for other provisioners?

And also if we want to have this, these configs should get passed into the configs of the node plugin and controller, and we should also change the Get capabilities RPC response accordingly, So operations will not get panicked and K8s will know that we don't have these functionalities...

mhkarimi1383:
With that we could have some experimental features to be able to be enabled or disabled, and it will make an infra for that

mhkarimi1383:
And I think we can just do a flag for enabling snapshots or resize, so no matter how we handle disabling in our side, the user should just tell that I don't want or I want this feature

tiagolobocastro:
> @tiagolobocastro do we have this functionality for other provisioners?

We don't, though I'm fine with it
 
> And also if we want to have this, these configs should get passed into the configs of the node plugin and controller, and we should also change the Get capabilities RPC response accordingly, So operations will not get panicked and K8s will know that we don't have these functionalities...

Good catch, yes we might have to change the csi capabilities which we report...

mhkarimi1383:
My suggestion is that we will have feature list in the code, and we will have a function called 

is_feature_enabled

And we will check it in the first step in our RPC handler

mhkarimi1383:
> My suggestion is that we will have feature list in the code, and we will have a function called 
> 
> is_feature_enabled
> 
> And we will check it in the first step in our RPC handler


And will check the same when reporting capabilities

deploy/helm/rawfile-localpv/values.yaml
pasha-gurkov: I don't believe there's any use-case, but I trust your judgement.

mhkarimi1383: I'm more Okay with the feature flag model that I said
Harder to implement, but better for later use cases and easier for users

pasha-gurkov:
@mhkarimi1383 @tiagolobocastro I've found these bits reporting the capabilities:
 - `GetPluginCapabilities()`
 - `NodeGetCapabilities()`
 - `ControllerGetCapabilities()`
 
all in `rawfile_servicer.py`. Is there anything else I'm missing?

Also, how'd you like this to be implemented? What about:
- having a `capabilities` top level key in the chart values, like `capabilities.snapshots.enabled` and `capabilities.resize.enabled`
- turning off containers if those are false
- passing these as environment variables to controller/node-plugin and modifying the `GetCapabilities()` calls accordingly

Also, judging by what I read in `VolumeManager`, snapshotting is required for volume cloning, right?

mhkarimi1383:
Yeah snapshotting is required for cloning

mhkarimi1383:
Yeah that's great it's better than having a key pair of features and a boolean value, with this we can make some features configurable later on

Also note that we should validate that feature is enabled or not in gRPC requests too

pasha-gurkov:
@mhkarimi1383 does this look okay to you?

mhkarimi1383:
Yeah, just one small thing, please move capabilities in settings model

Something like

```python
class BaseCapability(BaseModel):
    enabled: bool

class SnapshotCapability(BaseCapability):
    ...
    # we can add extra configs per capacity

class ResizeCapability(BaseCapability):
    ...
    # we can add extra configs per capacity

class Capabilities(BaseModel):
    resize: ResizeCapability
    snapshot: SnapshotCapability
```

in add a key called capabilities with Capabilities type in driver settings

And envs will be with nasted model with (__)
like

CSI_DRIVER__CAPABILITIES__RESIZE__ENABLED

It's just more strict and extendable 

But overally everything is great, thanks

pasha-gurkov:
@mhkarimi1383 thank you for the feedback, I'll push it next week!

pasha-gurkov:
@mhkarimi1383 pushed it && rebased on top of develop, could you confirm if that's what you're looking for?

mhkarimi1383 (approved):

mhkarimi1383:
Thanks looks Ok to me, we should also be aware that we have to not forget that we need statefulset for the API server too

pasha-gurkov:
Cool, @mhkarimi1383 do you want me to squash here as well?

tiagolobocastro (approved):

mhkarimi1383:
@pasha-gurkov,

Yeah, please squash your commits you can also use our good friend MergifyIO

pasha-gurkov:
@mhkarimi1383 Squashed and rebased on top of develop!

mhkarimi1383:
@mergify queue

PR #291, opened 2026-02-11
feat: Improve overall configuration flexibility

pasha-gurkov:
Add hostNetwork switch for the node component

Allow customisation for healthz driver-registrar port

Add topology knobs (affinity, nodeSelector)

Drop csi-probe port

Rationale: hostNetwork is valuable if the IP address space is scarce so we should support it; for that to work, every port should be customisable; topology knobs allow to choose where to run with finesse; and finally `grep 9808` doesn't give anything in the repo but the chart itself, so I assume it's an unused remnant (from 9bafb101ac00ba6aa3fe6504f745a15886fee677).

### DIFF
--- a/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
+++ b/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
@@ -17,6 +17,10 @@ spec:
       priorityClassName: {{ .Values.controller.priorityClassName }}
       tolerations:
         {{- .Values.controller.tolerations | toYaml | nindent 8 }}
+      nodeSelector:
+        {{- .Values.controller.nodeSelector | toYaml | nindent 8 }}
+      affinity:
+        {{- .Values.controller.affinity | toYaml | nindent 8 }}
       volumes:
         - name: socket-dir
           emptyDir: {}
@@ -83,9 +87,6 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
-          ports:
-            - name: csi-probe
-              containerPort: 9808
           resources:
             {{- include "rawfile-localpv.controller-resources" . | nindent 12 }}
         - name: external-resizer
--- a/deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
+++ b/deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
@@ -16,8 +16,15 @@ spec:
     spec:
       serviceAccount: {{ include "rawfile-localpv.fullname" . }}-driver
       priorityClassName: {{ .Values.node.priorityClassName }}
+      {{- if .Values.node.hostNetwork }}
+      hostNetwork: true
+      {{- end }}
       tolerations:
         {{- .Values.node.tolerations | toYaml | nindent 8 }}
+      nodeSelector:
+        {{- .Values.node.nodeSelector | toYaml | nindent 8 }}
+      affinity:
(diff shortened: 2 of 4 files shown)


deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
```diff
@@ -13,10 +13,14 @@ spec:
...
-      serviceAccount: {{ include "rawfile-localpv.fullname" . }}-driver
-      priorityClassName: {{ .Values.controller.priorityClassName }}
+      nodeSelector:
+        {{- .Values.controller.nodeSelector | toYaml | nindent 8 }}
       tolerations:
         {{- .Values.controller.tolerations | toYaml | nindent 8 }}
+      affinity:
+        {{- .Values.controller.affinity | toYaml | nindent 8 }}
+      serviceAccount: {{ include "rawfile-localpv.fullname" . }}-driver
+      priorityClassName: {{ .Values.controller.priorityClassName }}
```
tiagolobocastro: nit: can we keep these on the top to avoid diff change

deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
```diff
@@ -14,10 +14,17 @@ spec:
...
+      {{- if .Values.node.hostNetworking }}
+      hostNetwork: true
+      {{- end }}
+      nodeSelector:
+        {{- .Values.node.nodeSelector | toYaml | nindent 8 }}
       tolerations:
         {{- .Values.node.tolerations | toYaml | nindent 8 }}
+      affinity:
+        {{- .Values.node.affinity | toYaml | nindent 8 }}
+      serviceAccount: {{ include "rawfile-localpv.fullname" . }}-driver
```
tiagolobocastro: same here

deploy/helm/rawfile-localpv/values.yaml
```diff
@@ -95,6 +101,9 @@ node:
     # -- Overrides default image pull policy for node component
     pullPolicy: ""
 
+  # -- Enables hostNetwork for node component
+  hostNetworking: false
```
tiagolobocastro: nit: name it the same?

```suggestion
  hostNetwork: false
```

tiagolobocastro (approved):

deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
pasha-gurkov: Sure thing, modified!

deploy/helm/rawfile-localpv/templates/node-plugin/daemonset.yaml
pasha-gurkov: :+1:

deploy/helm/rawfile-localpv/values.yaml
pasha-gurkov: Sure -- I looked at a couple of other charts to mimic them; if you think naming those the same is better, let's do that instead.

pasha-gurkov:
@tiagolobocastro All addressed!

mhkarimi1383 (approved):

mhkarimi1383:
@pasha-gurkov thanks, can you please squash

pasha-gurkov:
@mhkarimi1383 squashed into one.

mhkarimi1383:
@mergify queue

PR #306, opened 2026-03-06
fix: Don't count temporary dir as a snapshot

pasha-gurkov:
When deleting a volume, the snapshots are counted, and due to a logical error, a
temp dir is counted as a snapshot as well. The result is, the volume gets deleted, but
the directories linger, because the driver thinks there are live snapshots. This fixes it.


tiagolobocastro (approved):
oh, nice find!

pasha-gurkov:
@mergify queue

PR #307, opened 2026-03-06
fix: Reference correct storage pool on CreateVolume

pasha-gurkov:
Fix always referencing the default storage pool to a correct one.

Kudos to Tiago Castro <tiagolobocastro@gmail.com> for noticing.


tiagolobocastro (approved):

pasha-gurkov:
@mergify queue

PR #308, opened 2026-03-06
Several improvements on the task manager [tackles flaky CI test]

tiagolobocastro:
test: log snapshot data with command
    
    Don't we also need to check if the expected data is correct???
    
---

    fix: store in-memory task before submission
    
    Otherwise we may get callback with no task stored.
    Also cater for this by logging an error and exiting if we ever
    encounter this scenario again as this could crash the task_manager.
    
---

    feat: enrich task information with timestamps and error
    
    Adds created, saved and retried timestmap - this should make it easier
    to understand the state of a task.
    
    Also added last error to the task, so we can find out why
    a task was saved as failed.
    
---

    fix: handle crash of the TaskManager retry worker
    
    In case there's errors during the task manager worker, then
    we catch this error, and log it.
    
    Since we don't run the tasks directly (we only schedule them), we
    should not be facing any errors here other than those related to
    encode/decode of the tasks.json and any I/O errors reading/writing it.
    
    In the future we may want to trigger a container graceful shutdown, thus
    making this error more visible.
    Alternatively, we could expose this via prometheus metrics.

---

    chore: log all the volumes that require GC
    
    Currently we don't seem to be GCing all volumes.
    We now always run the GC with dryRun so we can at least log
    all the volumes at node plugin startup.

---

    style: don't log attached loops query
    
    This ends up flooding the log and it's probably not that useful now
    If we add a trace level we could make it a trace log
    
---

    dev: git ignore .vscode dir

### DIFF
--- a/rawfile/tests/test_smoke.py
+++ b/rawfile/tests/test_smoke.py
@@ -769,7 +769,8 @@ def _(restore_pvc, restore_pod):
         stdout=True,
         tty=False,
     )
-    logger.error(f"C: {out}")
+    # todo: aren't we missing the check for correct data here!?
+    logger.info(f"cat /{pvc_name}/file => {out}")
 
 
 @when("we delete the snapshot")
--- a/rawfile/utils/task_manager.py
+++ b/rawfile/utils/task_manager.py
@@ -72,14 +72,22 @@ class TaskManager:
 
     def retry_worker(self):
         while not self._shutting_down:
-            self.retry_all()
-            for task_id, _ in self.get_tasks(retriable=False).items():
-                self.remove_task(task_id)
-            for task_id, _ in self.get_tasks(state=TaskState.COMPLETED).items():
-                self.remove_task(task_id)
+            try:
+                self.retry_all()
+                for task_id, _ in self.get_tasks(retriable=False).items():
+                    self.remove_task(task_id)
+                for task_id, _ in self.get_tasks(state=TaskState.COMPLETED).items():
+                    self.remove_task(task_id)
+            except Exception as e:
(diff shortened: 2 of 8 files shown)


rawfile/utils/task_manager.py
```diff
@@ -71,20 +77,28 @@ def __init__(
...
+                self.retry_all()
+                for task_id, _ in self.get_tasks(retriable=False).items():
+                    self.remove_task(task_id)
+                for task_id, _ in self.get_tasks(state=TaskState.COMPLETED).items():
+                    self.remove_task(task_id)
+                time.sleep(self._retry_interval)
+            logger.info("TaskManager Worker exit after shutdown request...")
+        except Exception as e:
+            logger.error("TaskManager Worker failed with an exception", exc=e)
+            # How should we handle this? Tasks will get stuck...
```
tiagolobocastro: How should we handle this?

rawfile/tests/test_smoke.py
```diff
@@ -769,7 +769,8 @@ def _(restore_pvc, restore_pod):
         stdout=True,
         tty=False,
     )
-    logger.error(f"C: {out}")
+    # todo: aren't we missing the check for correct data here!?
```
pasha-gurkov, line 772: We are, yeah :)

pasha-gurkov (approved):

rawfile/tests/test_smoke.py
tiagolobocastro: Created https://github.com/openebs/rawfile-localpv/issues/310 to track this

tiagolobocastro:
@mergify queue

PR #313, opened 2026-03-20
fix: GetCapacity calls for freshly created volumes

pasha-gurkov:
Fix the behaviour where a freshly created volume may not affect the available
capacity leading to unwanted overprovisioning.

CreateVolume -> GetCapacity -> NodeStageVolume call sequence means that
effectively the volume won't be formatted or mounted at the time GetCapacity is
called. Thus the proposed change: volume_stats() are still returned even if the
volume is not yet attached as a loopdev or formatted. The only stat depending on
it is used space in terms of filesystem, and it's used only in metrics now.

That opens the possibility to create a lot of volumes in a short amount of time,
filling up the whole disk, effectively ignoring `reserved_capacity`, triggering
kubelet DiskPressure and evicting pods from the node (ask me how I know :) )

### DIFF
--- a/rawfile/utils/volume_manager.py
+++ b/rawfile/utils/volume_manager.py
@@ -263,14 +263,21 @@ class VolumeManager:
             pool = meta.get("storage_pool", config.csi_driver.default_pool)
             thin_provision = meta.get("thin_provision", False)
             file = img_file(volume_id=volume_id)
+
+            try:
+                file_stats = stat(file)
+            except FileNotFoundError:
+                return None
+
             loop_devs = attached_loops(file.as_posix())
-            if not (loop_devs and len(loop_devs)):
-                return None
-            mountpoint = device_to_mountpoint(loop_devs[0])
-            if not mountpoint:
-                return None
-            filesystem_stats = statvfs(mountpoint)
-            file_stats = stat(file)
+            if loop_devs and len(loop_devs):
+                mountpoint = device_to_mountpoint(loop_devs[0])
+                if mountpoint:
+                    filesystem_stats = statvfs(mountpoint)
+                else:
+                    filesystem_stats = {"fs_usage": 0}
+            else:
+                filesystem_stats = {"fs_usage": 0}
 
             # note that "sparse" here might differ from metadata's
             # "thin_provision": if the volume was created as thick but formatted


rawfile/utils/volume_manager.py
```diff
@@ -263,14 +263,21 @@ def get_volume_stats(self, volume_id) -> VolumeStats | None:
...
             file = img_file(volume_id=volume_id)
-            loop_devs = attached_loops(file.as_posix())
-            if not (loop_devs and len(loop_devs)):
-                return None
-            mountpoint = device_to_mountpoint(loop_devs[0])
-            if not mountpoint:
+
+            try:
+                file_stats = stat(file)
+            except FileExistsError:
```
tiagolobocastro: Did you mean to use `FileNotFoundError` ?

rawfile/utils/volume_manager.py
```diff
@@ -263,14 +263,21 @@ def get_volume_stats(self, volume_id) -> VolumeStats | None:
...
+
+            loop_devs = attached_loops(file.as_posix())
+            if loop_devs and len(loop_devs):
+                mountpoint = device_to_mountpoint(loop_devs[0])
+                if mountpoint:
+                    filesystem_stats = statvfs(mountpoint)
+                else:
+                    filesystem_stats = {"fs_usage": 0}
+            else:
+                filesystem_stats = {"fs_usage": 0}
```
tiagolobocastro, line 280: This is just to complement the stats right, the main change is that we no longer return None if the file is not mounted?
Then we can use the `physical_size` ?

rawfile/utils/volume_manager.py
pasha-gurkov: That's a great catch, yes!

rawfile/utils/volume_manager.py
pasha-gurkov: - correct, and the `fs_usage` is basically only used in `metrics.py`
- correct, that's what we're using to calculate the capacity

rawfile/utils/volume_manager.py
pasha-gurkov: Changed.

rawfile/utils/volume_manager.py
tiagolobocastro: cool, thank you

tiagolobocastro (approved):
I wonder if as a future improvement, we could record `fs_usage` when unstaging volumes, thus allowing us to report `fs_usage` when not mounted.

pasha-gurkov:
@mergify queue

pasha-gurkov:
> I wonder if as a future improvement, we could record `fs_usage` when unstaging volumes, thus allowing us to report `fs_usage` when not mounted.

Hm, I guess adding that, and whether the volume is staged or not to metadata? `is_staged` or similar might be useful when reporting metrics I guess, to give a more comprehensive picture.

pasha-gurkov:
@tiagolobocastro created https://github.com/openebs/rawfile-localpv/issues/321, please check if that's what you meant.

tiagolobocastro:
> @tiagolobocastro created #321, please check if that's what you meant.

Yes that was it, thank you
The `is_staged` is also a good one!

PR #316, opened 2026-03-25
ci: update setup trivy version

pchandra19:
The setup trivy action was compromised. See : https://github.com/aquasecurity/setup-trivy/issues/31
The older versions (releases/tags) were removed from the https://github.com/aquasecurity/setup-trivy/releases

I tried dependabot to update but it says that the package is no longer vulnerable and hence it does not update, hence the manual PR.

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           driver-opts: |
             network=host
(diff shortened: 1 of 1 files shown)


tiagolobocastro (approved):

tiagolobocastro:
@mergify queue

PR #337, opened 2026-05-13
fix(metrics): rawfile_pool_capacity metrics semantic

luoxingao0930:
`storage_pool.py:49-50` defines the pool's capacity as:

```python
# reserved_capacity is the capacity reserved for everything else but this pool
# so the capacity reserved for this pool is (disk_total_size - reserved_capacity)
```

But `rawfile_pool_capacity_bytes` metric was exposing `statvfs.fs_size` directly — the FS size, not the pool's allocated capacity. Thus fixing the metric calculation and exposing an additional rawfile_pool_backing_fs_capacity_bytes for the backing filesystem size.


pasha-gurkov (approved):

pasha-gurkov:
@mergify queue

PR #347, opened 2026-06-08
fix(taskmanager): fix taskmanager's task reader default retry_count values

pasha-gurkov:
Some tasks don't have retry_count at all, so when reading those we might get errors accessing "retry_count" key if it's not there. This commit fixes it.


mhkarimi1383 (approved):

pasha-gurkov:
@mergify queue

PR #348, opened 2026-06-08
feat(chart): Add capacityPollInterval for externalProvisioner

pasha-gurkov:
Add capacityPollInterval to Helm chart values, parameter determining how long the external-provisioner waits before checking for storage capacity changes.


tiagolobocastro (approved):
So it's not refreshing after every volume operation?

Also `--capacity-for-immediate-binding` seems interesting

pasha-gurkov:
> So it's not refreshing after every volume operation?

AFAIU it does; but it's a race. It _seems_ it's better to be able to control this aspect that not, but I'm unsure if that brings any real positives.

> Also `--capacity-for-immediate-binding` seems interesting

Not totally sure what you mean, seems in that case capacity isn't used for scheduling (as the volume already exists at the time of scheduling). Maybe for monitoring based on CSIStorageCapacity objects?

pasha-gurkov:
@mergify queue

PR #281, opened 2026-02-01
:sparkles: API Server Added

mhkarimi1383:

PR #351, opened 2026-06-09
fix: imageurl rendered without quotes

krishnaGajabi:
This PR fixes the issue>>
When a registry value contains YAML-special characters (e.g. {, }, [, ], :, etc.), the rendered YAML is invalid because the parser interprets those characters as flow indicators rather than literal string content.

### DIFF
--- a/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
+++ b/deploy/helm/rawfile-localpv/templates/controller/statefulset.yaml
@@ -27,7 +27,7 @@ spec:
           emptyDir: {}
       containers:
         - name: csi-driver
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait (approved):

mhkarimi1383:
This will conflict with the API Server PR (#281) due to the deployment-related filename change. Options are: wait for that PR to merge, align API Server PR with the same change, or merge now if required and resolve the conflict later.

mhkarimi1383 (approved):

tiagolobocastro (approved):

tiagolobocastro:
@mergify queue

tiagolobocastro:
> This will conflict with the API Server PR (#281) due to the deployment-related filename change. Options are: wait for that PR to merge, align API Server PR with the same change, or merge now if required and resolve the conflict later.

Since we're having a release, let's merge it now so we can pick it up in the umbrella chart

PR #353, opened 2026-06-11
:recycle: Use CoW on snapshot restore when possible

### LINKED ISSUE #309: Creating Volumes from Snapshots should be (almost) instant
Correct me if I'm wrong, but shouldn't Volume creation be (almost) instant when they are created from Snapshots (with COW supported)? 

Because now I can't use Velero to create PVC Backups (Velero creates PVC referencing Snapshot), as it copies the Snapshot image (created instantly) byte-for-byte over to a new image file, which just takes too long. I thought it would just do `cp --reflink=always snap.img restored.img`.

See: https://github.com/openebs/rawfile-localpv/blob/0593b8947349a335ff092d461bfb0a4e72b73cb6/rawfile/utils/snapshot_manager.py#L123-L136
Used here: https://github.com/openebs/rawfile-localpv/blob/0593b8947349a335ff092d461bfb0a4e72b73cb6/rawfile/utils/volume_manager.py#L148-L153

mhkarimi1383:
Uses CoW when CoW is possible between source and destination also supports cloning between multiple pools

Resolves #309

### DIFF
--- a/rawfile/utils/snapshot_manager.py
+++ b/rawfile/utils/snapshot_manager.py
@@ -1,7 +1,8 @@
-from os import fsync
+import os
 from pathlib import Path
 import time
 from dataclasses import dataclass
+import consts
 from utils.commands import run
 from config import config
 from utils.errors import FsFreezeNotSupportedOnBlockVolumes, SnapshotCreateVolumeInUse
@@ -9,8 +10,8 @@ from utils.lock import VolLock
 from utils.rawfile import (
     attached_loops,
     img_file,
+    is_cow_supported,
     metadata,
-    patch_metadata,
     snapshots_dir,
 )
 from glob import glob
@@ -72,15 +73,6 @@ class SnapshotManager:
 
                 creation_time = time.time()
                 Path(f"{snap_path}.creating").unlink(missing_ok=True)
-                meta = metadata(volume_id)
-                reflink_attached = list(set(meta.get("reflink_attached", [])))
-                if copy_on_write:
-                    reflink_attached.append(name)
-                patch_metadata(
-                    volume_id,
-                    meta.get("storage_pool", config.csi_driver.default_pool),
-                    {"reflink_attached": reflink_attached},
-                )
                 return Snapshot(
                     name=name,
                     volume_id=volume_id,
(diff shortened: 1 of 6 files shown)


rawfile/utils/volume_manager.py
```diff
@@ -181,14 +181,7 @@ def rmdir(path: Path):
...
-            logger.warning(
-                "Volume has COW Snapshots attached, skipping destroy, will be destoyed when all snapshots are removed",
-                volume_id=volume_id,
-                snapshots=total_snapshots,
-            )
-            return
-        elif total_snapshots > 0:
+        if total_snapshots > 0:
             logger.warning(
                 "Volume has Snapshots(without COW) attached, will only remove volume data",
```
cometship: I think the `(without COW)` is (sometimes) incorrect now

mhkarimi1383: Yeah, but mentioning it would help users to understand the reason

mhkarimi1383:
I have to also remove reflink attach counter from volume meta, I forgot that

rawfile/utils/volume_manager.py
mhkarimi1383: Sorry I was wrong, since we are talking about all of the snapshots...

mhkarimi1383: Fixed, The message

rawfile/utils/snapshot_manager.py
```diff
@@ -110,30 +102,26 @@ def delete_snapshot(self, volume_id: str, name: str, temporary: bool = False):
...
+        copy_on_write = (
+            copy_on_write_param
+            if copy_on_write_param is not None
+            else consts.COW_SUPPORT_MAP.get(
+                volume_meta.get("storage_pool", None), False
+            )
+        ) and is_cow_supported(Path(os.path.dirname(snap_path)), destination)
+        reflink = "always" if copy_on_write else "never"
+        cmd = (
+            f"cp --sparse=auto --reflink={reflink} {snap_path} {destination.as_posix()}"
```
tiagolobocastro, line 122: Do we not need a flush if not using cow?

rawfile/utils/volume_manager.py
```diff
@@ -181,16 +181,9 @@ def rmdir(path: Path):
         temp_snapshots = list(snapshots_dir(volume_id, temporary=True).glob("*"))
         total_snapshots = len(snapshots) + len(temp_snapshots)
         meta = metadata_or(volume_id)
-        if len(meta.get("reflink_attached", [])) > 0:
+        if total_snapshots > 0:
```
tiagolobocastro, line 184: is there 2 fixes in this PR? seems it's also removing the reflinks and making decision based on the dynamically discovered snapshots?

mhkarimi1383: This PR totally removes the connection and dependency between CoW files with their source
Two actions was needed for that, removing reflink attached and use CoW for clones to have instant restores

rawfile/utils/snapshot_manager.py
mhkarimi1383: We are not doing it in other places after cp, I think it will handle that by itself.

tiagolobocastro: Are there other places where we do cp without reflink? Then those are wrong as well, cp does not flush/sync AFAIK

mhkarimi1383: Yeah we have, We have to change them too...
I think let's merge this and I will open another PR for That fix.

tiagolobocastro (approved):

mhkarimi1383:
@mergify queue

mhkarimi1383:
@mergify squash

PR #372, opened 2026-07-17
feat(analytics): make GA analytics DNS nameservers configurable

krishnaGajabi:
## What

Makes the DNS nameservers used to resolve the Google Analytics collection
endpoint configurable, instead of the previously hardcoded `8.8.8.8`. Also
fixes several latent bugs in the analytics `DNSAdapter`.

## Why

The GA4 client resolved `www.google-analytics.com` through a custom adapter
that hardcoded `8.8.8.8` as the resolver. This is problematic:

- In clusters that restrict egress to public DNS, `:53` to `8.8.8.8` is
  blocked, so analytics fails even though cluster CoreDNS (→ upstream) would
  have resolved it fine.
- It bypasses cluster CoreDNS entirely.

It also modernizes a deprecated dnspython call — `Resolver.query()` (deprecated
since dnspython 2.0 in favor of `resolve()`) — and fixes a real bug: bare `raise`
statements used outside any `except` block, which raise `RuntimeError: No active
exception to re-raise` instead of a meaningful error. These are now explicit
exceptions.

## Changes

- **Configurable nameservers** via new `ga_dns` config (env `GA_DNS`). Empty
  (the code default) means "use the pod/system resolver" (e.g. CoreDNS). Accepts
  a comma-separated string or list; each entry may include a port, e.g.
  `8.8.8.8:53` or `[2001:4860:4860::8888]:53`.
- **`utils/net.py`**: parses/validates `ip[:port]` nameservers (IPv4/IPv6),
  dropping invalid entries with a warning rather than failing hard. Uses
  `NoDecode` + a before-validator so `GA_DNS=8.8.8.8:53` isn't JSON-decoded by
  pydantic-settings before validation.
- **`ga4/client.py`**: builds a per-server `Do53Nameserver` (dnspython's
  `nameservers` list can't carry a port); only installs the DNS adapter when
  `ga_dns` is set. Fixes the `query()`→`resolve()` call and replaces bare
  `raise` with explicit exceptions.
- **Helm**: adds `analytics.gaDnsNameservers` (plus a `global` override) and
  wires `GA_DNS` into the controller StatefulSet and node DaemonSet; regenerates
  the chart README.

## Backward compatibility

The Helm chart default is `analytics.gaDnsNameservers: "8.8.8.8:53"`, so existing
installs keep the previous behavior. Operators can set it to `""` to use the
cluster's default resolver (CoreDNS), or point it at a specific resolver.

## Testing

- Added unit tests in `rawfile/tests/test_net.py` covering `split_host_port`
  and `parse_nameservers` (IPv4/IPv6, ports, comma-strings, lists, empty →
  disabled, invalid → dropped-with-warning, and a `NoDecode` regression guard).
- `pre-commit run --all-files` clean.

## Notes for reviewers

- IPv6 support is for the **nameserver address** only; endpoint resolution is
  A-record (IPv4).

### DIFF
--- a/deploy/helm/rawfile-localpv/README.md
+++ b/deploy/helm/rawfile-localpv/README.md
@@ -27,6 +27,7 @@ Please follow the [install guide](https://github.com/openebs/rawfile-localpv/tre
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | analytics.enabled | bool | `true` | Enable OpenEBS analytics which help track engine traction and usage. |
+| analytics.gaDnsNameservers | string | `"8.8.8.8:53"` | Comma-separated DNS nameservers (each optionally `ip:port`) used to resolve the analytics endpoint. Leave empty to use the cluster's default resolver (CoreDNS). |
 | auth.enabled | bool | `true` | Enables authentication for internal gRPC server |
 | auth.secretName | string | `""` | If managing secrets outside the chart, use this to reference the secret name; otherwise, leave empty. |
 | auth.token | string | `""` | Sets authentication token for internal gRPC server, will generate one if nothing provided |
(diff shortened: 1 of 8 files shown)


mhkarimi1383 (approved):

mhkarimi1383:
@mergify rebase

mhkarimi1383:
@mergify queue

PR #375, opened 2026-07-17
:bug: Use findmnt with more stable output instead of using mount

### LINKED ISSUE #370: Volume expansion fails: `NodeExpandVolume` runs `losetup -c` on the backing file instead of the loop device
## Summary

Online volume expansion (`NodeExpandVolume`) fails on some environments (reproduced on
Ubuntu 22.04 / kernel 5.15; not reproduced on Ubuntu 24.04). The driver determines the device
to resize by scraping `mount(8)` output. On the affected environment `mount(8)` reports the
loop mount's **backing file** (`.../disk.img`) instead of its **loop device** (`/dev/loopN`),
so the driver runs `losetup -c /path/to/disk.img`, which fails — `losetup -c` requires a loop
device. The RPC aborts before the filesystem is grown, so the PVC never reaches its requested
size.

The underlying problem is that the code treats `mount(8)` human-readable output as a stable
"mountpoint -> device" API. It is not: `mount(8)` legitimately presents a loop mount by its
backing file in some environments. The kernel's `/proc/self/mountinfo` always records the
authoritative `/dev/loopN`, but the code doesn't read it.

## Actual behavior

The PVC stays at its original capacity. PVC status shows:

```
type: NodeResizeError
message: failed to expand pvc with Expander.NodeExpand failed to expand the volume :
  rpc error: code = Unknown desc = Exception calling application:
  Command 'losetup -c /var/local/openebs/rawfile/default-pool/pvc-<id>/disk.img'
  returned non-zero exit status 1.
```

The node plugin logs the same `losetup -c .../disk.img` `CalledProcessError`.

## Expected behavior

The PVC (and the pod's filesystem) expands to the requested size.

## Root cause

The volume is correctly staged on a loop device — the driver mounts `/dev/loopN`, and the
kernel records it: `findmnt -no SOURCE <globalmount>/mount` returns `/dev/loop42`.

`NodeExpandVolume` refreshes the loop device capacity with:

```python
# rawfile/rawfile_servicer.py  (RawFile node servicer)
volume_path = Path(request.volume_path).resolve()
run(f"losetup -c {volume_path}")   # -c requires a /dev/loopN, not a file
```

`volume_path` comes from `get_device_for_mountpoint`, which derives the device by parsing
`mount(8)` output and taking the first field of the matching line:

```python
# rawfile/filesystem/utils.py
output = run(f"mount | grep 'on {mountpoint}'", check=True, capture_output=True)
...
return Path(lines[0].split(" ")[0].strip()).resolve().as_posix()
```

On the affected environment, that first field is the backing file `.../disk.img` (even though
the kernel's mount source is `/dev/loop42`), so the driver runs `losetup -c .../disk.img`,
which fails. Confirmed standalone that `losetup -c` rejects a file:

```
# losetup -c /dev/loop0        -> exit 0
# losetup -c /tmp/disk.img     -> losetup: /tmp/disk.img: failed to use device: No such device ; exit 1
```

Because `losetup -c` raises, `NodeExpandVolume` returns an error before the filesystem-resize
step (`fs.resize()` in `bd2fs.py`), so nothing grows.

The key defect is depending on `mount(8)` presentation output for a device path. The
authoritative device is always in `/proc/self/mountinfo` (`findmnt -no SOURCE` returns
`/dev/loopN`), regardless of environment. This mount-parsing + `losetup -c` shape dates back to
the original `bd2fs` extraction (commit c58dd14, 2019) and has never been revisited.

## Suggested fix

Stop depending on `mount(8)` human-readable output. Either:

1. **Preferred:** have `get_device_for_mountpoint` read the real source device from
   `/proc/self/mountinfo` (always `/dev/loopN`) instead of scraping `mount`. Fixes all callers
   and removes the fragile unanchored `grep`.
2. Or, in the expand path, translate the backing file to its loop device via `losetup -j <file>`
   (the existing `attached_loops()` helper) before calling `losetup -c`.

Option 1 is behavior-preserving where `mount(8)` already returns the device, and fixes the
environments where it returns the backing file.

### Which version of rawfile are you using?

0.14.0

### What did you expect to happen?

Volume Should Be Expanded

## Steps to reproduce

1. Deploy rawfile-localpv on an affected host (e.g. Ubuntu 22.04 / kernel 5.15).
2. Create a `Filesystem` PVC (ext4/btrfs) and mount it in a pod.
3. Edit the PVC to request a larger size (e.g. 512Mi -> 1Gi).
4. Start a pod so node expansion runs.

## Environment

- rawfile-localpv: `<chart / image version, e.g. v0.14.0>`
- Kubernetes: v1.33.1 (kind)
- **Reproduces:** Ubuntu 22.04 host, kernel 5.15.0-138-generic
- **Not reproduced:** Ubuntu 24.04 host (kernel 6.8) — `mount(8)` there shows `/dev/loopN`, so
  `get_device_for_mountpoint` returns a loop device and expansion succeeds. This likely explains
  why CI (Ubuntu 24.04) is green.

### The output of the following commands will help us better understand what's going on

```
type: NodeResizeError
message: failed to expand pvc with Expander.NodeExpand failed to expand the volume :
  rpc error: code = Unknown desc = Exception calling application:
  Command 'losetup -c /var/local/openebs/rawfile/default-pool/pvc-<id>/disk.img'
  returned non-zero exit status 1.
```

mhkarimi1383:
Resolves #370

Since I have a much newer kernel, Tested with

```
> truncate -s 10GiB ./file
> mkfs.xfs ./file
 meta-data=./file                 isize=512    agcount=4, agsize=655360 blks
          =                       sectsz=512   attr=2, projid32bit=1
          =                       crc=1        finobt=1, sparse=1, rmapbt=1
          =                       reflink=1    bigtime=1 inobtcount=1 nrext64=1
          =                       exchange=1   metadir=0
 data     =                       bsize=4096   blocks=2621440, imaxpct=25
          =                       sunit=0      swidth=0 blks
 naming   =version 2              bsize=4096   ascii-ci=0, ftype=1, parent=1
 log      =internal log           bsize=4096   blocks=16384, version=2
          =                       sectsz=512   sunit=0 blks, lazy-count=1
 realtime =none                   extsz=4096   blocks=0, rtextents=0
          =                       rgcount=0    rgsize=0 extents
          =                       zoned=0      start=0 reserved=0
> mount ./file /mnt
> findmnt -no SOURCE --mountpoint /mnt
 /dev/loop0
```

### DIFF
--- a/rawfile/filesystem/utils.py
+++ b/rawfile/filesystem/utils.py
@@ -1,3 +1,4 @@
+import re
 import subprocess
 from pathlib import Path
 from utils.commands import run
@@ -14,13 +15,15 @@ def get_device_fs(device: str) -> str | None:
 def get_device_for_mountpoint(mountpoint: str) -> str | None:
     try:
         output = run(
-            f"mount | grep 'on {mountpoint}'",
+            f"findmnt -no SOURCE --mountpoint {mountpoint}",
             check=True,
             capture_output=True,
         )
     except subprocess.CalledProcessError:
         return None
-    lines = output.stdout.decode().strip().splitlines()
-    if not len(lines):
+    source = output.stdout.decode().strip()
+    if not source:
         return None
-    return Path(lines[0].split(" ")[0].strip()).resolve().as_posix()
+    # Strip bind-mount subpath notation, "/dev/loop1[/default]" -> "/dev/loop1"
+    device = re.sub(r"\[.*\]$", "", source)
+    return Path(device).resolve().as_posix()


tiagolobocastro (approved):

tiagolobocastro (changes requested):
failing tests

mhkarimi1383:
2026-07-17 13:40:19.145 | DEBUG    | utils.commands:run:43 - Shell command execution | {'check': True, 'capture_output': True, 'command': 'findmnt -no SOURCE --mountpoint /var/lib/kubelet/pods/d19be1c2-7b63-402a-90d3-5d7bf9506787/volumes/kubernetes.io~csi/pvc-261af500-cf7f-4253-9a27-4cbe675533a5/mount', 'start': datetime.datetime(2026, 7, 17, 13, 40, 19, 142930), 'returncode': 0, 'end': datetime.datetime(2026, 7, 17, 13, 40, 19, 145721), 'latency': datetime.timedelta(microseconds=2791), 'stderr': '', 'stdout': '/dev/loop1[/default]\n'}


I'm working on it.

mhkarimi1383:
Fixed.

tiagolobocastro (approved):

PR #416, opened 2026-09-08
feat: allow for node.podLabels and controller.podLabels to be set

### LINKED ISSUE #415: Allow for podLabels to be set on the DaemonSet Pods
### Describe the problem/challenge you have

Hello,

We have a need internally to set a non-podSelector label for the  rawfile-localpv DaemonSet Pods.

I am happy to create a PR that adds this in if you agree that this would be useful.

### Describe the solution you'd like

The ability to set node.podLabels and have them populated under spec.template.metadata.labels, similar to podAnnotations

### Anything else you would like to add?

_No response_

poblahblahblah:
This change allows for node.podLabels and controller.podLabels to be set in the Helm Chart's values.yaml file. This will allow users to configure Pod Labels that do not necessarily need to be Selector Labels

Closes #415


tiagolobocastro (approved):
Could you add it to the controller as well?

tiagolobocastro:
@poblahblahblah could you please squash the commits?

tiagolobocastro:
@poblahblahblah again merge commit has the same issue with DCO and also please don't add merge commits and instead do rebase to keep history clean

